PR #3, opened 2019-08-02
update secret key names for csi-driver

shivamerla:


storageprovider/storage_provider.go
```diff
@@ -10,12 +10,14 @@ import (
 )
 
 const (
-	arrayIPKey     = "arrayIp"
+	arrayIPKey     = "backend"
 	usernameKey    = "username"
 	passwordKey    = "password"
-	portKey        = "port"
+	portKey        = "servicePort"
```
rgcostea: Can you change the key names too?

shivamerla: updated other places too to use Backend instead of ArrayIP and ServicePort instead of Port. Added default port as well.

PR #5, opened 2019-08-06
Update CSP headers

rgcostea:
Changes to accept this header in the CSP are forthcoming in a separate review.


rgcostea:
> currently Token object expects array-ip..should we make it generic too?

Yes.  I missed that.  Thanks.

PR #14, opened 2019-08-08
add virtual device(vmdk) attach and detach support

shivamerla:


linux/virtual_device.go
```diff
@@ -0,0 +1,164 @@
...
+		}
+		if f.Mode()&os.ModeSymlink != 0 {
+			pathName, err := os.Readlink(diskBySerial + f.Name())
+			if err != nil {
+				util.LogError.Printf("unable to read symlink %s to get device details, err %s", f.Name(), err.Error())
+				return nil, err
+			}
+			device := &model.Device{}
+			// ../../sdb
+			devNames := strings.Split(pathName, "/")
```
raunakkumar: nit:  check size of devNames before dereferencing it

shivamerla: sure, will add this..

PR #20, opened 2019-08-10
CSI Driver (Clone Volume): Handle clone creation when size greater than source size

suneeth51:
- Volume clone creation with size >= snapshot size is allowed, else error will be returned.
- CSP handles the resize in CloneVolume() endpoint.

### DIFF
--- a/storageprovider/csp/container_storage_provider.go
+++ b/storageprovider/csp/container_storage_provider.go
@@ -243,8 +243,8 @@ func (provider *ContainerStorageProvider) CreateVolume(name, description string,
 
 // CloneVolume clones a volume on the CSP
 // nolint : gocyclo
-func (provider *ContainerStorageProvider) CloneVolume(name, description, sourceID, snapshotID string, opts map[string]interface{}) (*model.Volume, error) {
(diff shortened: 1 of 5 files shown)


suneeth51:
> nit: propagate the detail error back too

Sure, I will update it accordingly.

storageprovider/csp/container_storage_provider.go
```diff
@@ -243,8 +243,8 @@ func (provider *ContainerStorageProvider) CreateVolume(name, description string,
 
 // CloneVolume clones a volume on the CSP
 // nolint : gocyclo
-func (provider *ContainerStorageProvider) CloneVolume(name, description, sourceID, snapshotID string, opts map[string]interface{}) (*model.Volume, error) {
-	log.Tracef(">>>>> CloneVolume with name: %v, description: %v, sourceID: %v, snapshotID %v", name, description, sourceID, snapshotID)
+func (provider *ContainerStorageProvider) CloneVolume(name string, description string, resize int64, sourceID string, snapshotID string, opts map[string]interface{}) (*model.Volume, error) {
```
rgcostea: Why add string after name?  I would rename "resize" to "size" and put it after name, description, sourceID, snapshotID.

suneeth51: Agreed, I will update it as per the suggestion.

PR #24, opened 2019-08-14
Win DVP crash during mount request 

vidhutsingh:
Earlier code merge didn't go through to master . So creating new pull request. 
Regression is introduced where it is expecting to have State field in device struct , windows chapi doesnt return thus leading to plugin crash on Windows while running mount command.

### DIFF
--- a/chapi/chapiclient.go
+++ b/chapi/chapiclient.go
@@ -271,7 +271,7 @@ func (chapiClient *Client) AttachDevice(volumes []*model.Volume) (devices []*mod
 		log.Errorf("AttachDevice: Err:%s for volume(%s)", err.Error(), volumes[0].Name)
 		return nil, err
 	}
-	
+	// Win CHAPI doesnt return Device[0].State, added OR condition to handeled it.	
 	if len(devices) != 0 && ((devices[0].State == model.ActiveState.String()) || (devices[0].State == "")) {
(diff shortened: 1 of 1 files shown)


chapi/chapiclient.go
```diff
@@ -271,7 +271,7 @@ func (chapiClient *Client) AttachDevice(volumes []*model.Volume) (devices []*mod
 		log.Errorf("AttachDevice: Err:%s for volume(%s)", err.Error(), volumes[0].Name)
 		return nil, err
 	}
-	
+	// Win CHAPI doesnt return Device[0].State, added OR condition to handeled it.
```
rgcostea: So the only change is a comment?

 - fix doesn't
 - fix handled

vidhutsingh: Done. thanks.

chapi/chapiclient.go
```diff
@@ -271,7 +271,7 @@ func (chapiClient *Client) AttachDevice(volumes []*model.Volume) (devices []*mod
 		log.Errorf("AttachDevice: Err:%s for volume(%s)", err.Error(), volumes[0].Name)
 		return nil, err
 	}
-	
+	// Win CHAPI doesn't return Device[0].State, added OR condition to handele it.
```
rgcostea: handle

PR #29, opened 2019-08-19
deprecate util.logger and fix missing fatal logging with logrus ones

shivamerla:


logger/logger.go
```diff
@@ -392,6 +393,43 @@ func WithTime(t time.Time) *log.Entry {
 	return log.WithTime(t)
 }
 
+// HTTPLogger : wrapper for http logging
+func HTTPLogger(inner http.Handler, name string) http.Handler {
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		sourced().Infof(
+			">>>>> %s\t%s\t%s",
```
NimbleMike: The logrus package does not expand escape characters (e.g. \t).  Instead, it shows up like this in the log file:

`time="2019-08-19T11:13:11-07:00" level=info msg=">>>>> GET\t/api/v1/initiators\tHostInitiators" file="server.go:1947"`

For improved readability, we could change the format to something like this:

`">>>>> %s %s - %s",`

Generates output like this:

`time="2019-08-19T11:15:13-07:00" level=info msg=">>>>> GET /api/v1/initiators - HostInitiators" file="server.go:1947"`

Same comment for the logged exit event as well.  If you think the output file could be parsed, and the tab escape character is helpful, disregard comment.

Also, since entry/exit routines are typically trace logs, and not info logs, should these be sourced().Tracef()?

chapi/chapiclient.go
```diff
@@ -409,10 +411,10 @@ func (chapiClient *Client) GetDevices() (devices []*model.Device, err error) {
 	_, err = chapiClient.client.DoJSON(&connectivity.Request{Action: "GET", Path: devicesURI, Header: chapiClient.header, Payload: nil, Response: &chapiResp, ResponseError: &chapiResp})
 	if err != nil {
 		if errResp != nil {
-			util.LogError.Print(errResp.Info)
+			log.Error("GetDevices: error response", errResp.Info)
```
NimbleMike: Should this be the following to align with your other update (directly below).  Otherwise the error is going to append directly after "response".
`log.Error("GetDevices: error response, ", errResp.Info)`

shivamerla: actuall logger will add space in this case, but i agree we can keep it consistent.

logger/logger.go
shivamerla: sure, let me fix that..

chapi/chapiclient.go
NimbleMike: I don't see the above behavior.  Consider this line of code:
```log.Error("GetDevices: error response", "Hello World")```
It produces this in my log file:
```time="2019-08-19T11:54:29-07:00" level=error msg="GetDevices: error responseHello World" file="chapid_windows.go:83"```

PR #30, opened 2019-08-19
handler: Fix serial number usage

NimbleMike:
* Problem:
  * While working on porting the CHAPI client wrapper, noticed that GetDevices() and GetMounts() were not processing the request if a serial number was passed in.
* Implementation:
  * Remove the erroneous "return", after retrieving the serial number, in both the GetDevices() and GetMounts() endpoints
* Testing:
  * Tested GetDevices() API using CHAPI client wrapper that is in process
* Reviewers:
  * @shivamerla
* Bug: N/A


chapi2/chapiclient/chapiclient_linux.go
```diff
@@ -0,0 +1,201 @@
...
+
+	chapiClient, err = NewChapiSocketClientWithTimeout(timeout)
+	if err != nil {
+		return nil, err
+	}
+	return chapiClient, nil
+}
+
+// GetSerialNumber returns the volume serial number as per platform format
+func GetSerialNumber(serialNumber string) string {
```
shivamerla: we were thinking of dropping this logic from client side, so serial number for clients is always volume serial number. and linux CHAPI server will prepend "2" when checking for multipath device.

chapi2/chapiclient/chapiclient_windows.go
```diff
@@ -0,0 +1,299 @@
...
+// 	if err != nil && !strings.Contains(strings.ToLower(err.Error()), "file exists") {
+// 		return false, err
+// 	}
+// 	_, isDir, _ = util.FileExists(mountPoint)
+// 	if isDir == false {
+// 		log.Trace("Failed to created mountpoint" + mountPoint)
+// 		return false, errors.New("failed to create directory at mountpoint " + mountPoint)
+// 	}
+// 	return isDir, nil
+// }
```
shivamerla: can we add the binary file below to .gitignore.

chapi2/handler/handler.go
```diff
@@ -122,7 +122,6 @@ func GetDevices(w http.ResponseWriter, r *http.Request) {
 
 	if ok && len(keys[0]) > 0 {
 		serialNumber = keys[0]
-		return
```
shivamerla: good catch Mike, wondering how it used to work so far then?

NimbleMike:
Merging pull request.  Removed "in process" CHAPI client files.  Only file getting merged is handler.go change.

PR #33, opened 2019-08-21
chapi2: Add chapiclient package

NimbleMike:
* Problem:
  * Need to port CHAPI1 client package to CHAPI2
* Implementation:
  * handler_windows.go
    * Changed chapiPortFileName to ChapiPortFileName so that string is accessible to other packages (e.g. CHAPI client)
  * chapiclient.go
    * Ported CHAPI1 code to CHAPI2
    * NewChapiClient() and NewChapiClientWithTimeout() are platform independent methods to create a new CHAPI Client object
    * **Only** wrap CHAPI2 endpoints in this module
    * Simplified all endpoint handlers by wrapping chapiClient.client.DoJSON in a new chapiClient.chapiClientDoJSON() function (shared/common error handling)
    * Updated Response object to expect a cerrors.ChapiError object instead of a generic error
    * Compilation will fail if Client object does not support CHAPI2's Driver interface (i.e. keep parity between client and server)
  * chapiclient_linux.go
    * Provide Linux specific Client object creation methods
    * All Socket methods are in this Linux module since they're not applicable to Windows
  * chapiclient_windows.go
    * Provide Windows specific Client object creation methods
    * All HTTP port methods are in this Windows module since they're not applicable to Linux
  * chapiclient_util.go
    * Created a util stub module to add any support routines that are intended to extend the capabilities beyond the CHAPI endpoints
* Testing:
  * Successfully built Linux and Windows binaries
  * Created a chapid resident service for Windows and exercised available endpoints using chapiclient package
  * Linux testing and final implementation pending
* Reviewers:
  * @shivamerla, @suneeth51
* Bug: N/A

### DIFF
--- a/chapi2/chapiclient/chapiclient.go
+++ b/chapi2/chapiclient/chapiclient.go
@@ -124,7 +124,7 @@ func (chapiClient *Client) GetHostNetworks() (networks []*model.Network, err err
 
 	// Initialize CHAPI response object, submit request to specified endpoint, return status
 	chapiResp := Response{Data: &networks, Err: nil}
-	if _, err = chapiClient.client.DoJSON(&connectivity.Request{Action: "GET", Path: networksURI, Header: chapiClient.header, Payload: nil, Response: &chapiResp, ResponseError: &chapiResp}); err != nil {
+	if _, err = chapiClient.chapiClientDoJSON(&connectivity.Request{Action: "GET", Path: networksURI, Header: chapiClient.header, Payload: nil, Response: &chapiResp, ResponseError: &chapiResp}); err != nil {
 		return nil, err
 	}
 	return networks, nil
--- a/chapi2/chapiclient/chapiclient.go
+++ b/chapi2/chapiclient/chapiclient.go
@@ -1,0 +1,366 @@
+// (c) Copyright 2019 Hewlett Packard Enterprise Development LP
+
+package chapiclient
+
+import (
+	"fmt"
+	"strings"
+	"time"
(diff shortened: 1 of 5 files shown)


chapi2/chapiclient/chapiclient.go
```diff
@@ -0,0 +1,366 @@
...
+	// CHAPI request was successful; return the HTTP status code with no error
+	return statusCode, nil
+}
+
+// appendQuerySerialNumber appends a serial number query to the given URI
+func (chapiClient *Client) appendQuerySerialNumber(uri string, serialNumber string) string {
+	return chapiClient.appendQuery(uri, querySerialNumber, serialNumber)
+}
+
+// appendQuerySerialNumber appends a mount ID query to the given URI
```
shivamerla: nit: typo in function name..

chapi2/chapiclient/chapiclient.go
```diff
@@ -0,0 +1,366 @@
...
+func (chapiClient *Client) appendQueryMountPointID(uri string, serialNumber string) string {
```
shivamerla: mountID needs to be passed here?

chapi2/chapiclient/chapiclient.go
NimbleMike: Thanks for catching that.  Fixing it.

NimbleMike:
Fixed function name in comment plus function parameter name.  No functional changes, just corrected declaration and comment.

PR #34, opened 2019-08-23
logger:  Enhance sensitive keyword detection

NimbleMike:
* Problem:
  * util.IsSensitive() was returning false positives
  * Redundant lookup tables in util.IsSensitive() and logger.Scrubber()
* Implementation:
  * strings.go
    * Remove IsSensitive() function (moved to logger.go)
  * logger.go
    * Moved IsSensitive() function from strings.go into this module
      * Changed lookup table from string array to a map
      * Added "accesskey" to lookup table
    * Updated Scrubber() routine to use IsSensitive() function
  * client.go
    * Changed util.IsSensitive() to log.IsSensitive()
* Testing:
  * Built successfully for Windows and Linux
  * Tested on a Windows host
* Reviewers:
  * @shivamerla, @suneeth51, @rgcostea, @vidhutsingh , @raunakkumar 
* Bug: N/A

### DIFF
--- a/logger/logger.go
+++ b/logger/logger.go
@@ -419,17 +419,24 @@ func HTTPLogger(inner http.Handler, name string) http.Handler {
 // IsSensitive checks if the given key exists in the list of bad words (sensitive info)
 func IsSensitive(key string) bool {
 	// TODO: Add more sensitive words (lower-case) to this list
-	badWords := map[string]bool{
-		"x-auth-token": true,
-		"username":     true,
-		"user":         true,
-		"password":     true,
-		"passwd":       true,
-		"secret":       true,
-		"token":        true,
-		"accesskey":    true,
+	badWords := []string{
+		"x-auth-token",
+		"username",
+		"user",
+		"password",
+		"passwd",
+		"secret",
+		"token",
+		"accesskey",
 	}
-	return badWords[strings.ToLower(key)]
+	key = strings.ToLower(key)
+	for _, bad := range badWords {
+		// Perform case-insensitive and substring match
+		if strings.Contains(key, bad) {
+			return true
+		}
+	}
+	return false
 }
 
 // Scrubber checks if the args list contains any sensitive information like username/password/secret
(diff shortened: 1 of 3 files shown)


logger/logger.go
```diff
@@ -416,15 +416,28 @@ func HTTPLogger(inner http.Handler, name string) http.Handler {
...
+	// TODO: Add more sensitive words (lower-case) to this list
+	badWords := map[string]bool{
+		"x-auth-token": true,
+		"username":     true,
+		"user":         true,
+		"password":     true,
+		"passwd":       true,
+		"secret":       true,
+		"token":        true,
+		"accesskey":    true,
```
suneeth51: We should include 'accessSecret' as well?

logger/logger.go
```diff
@@ -416,15 +416,28 @@ func HTTPLogger(inner http.Handler, name string) http.Handler {
...
+	return badWords[strings.ToLower(key)]
+}
+
 // Scrubber checks if the args list contains any sensitive information like username/password/secret
 // If found, then returns masked string list, else returns the original input list unmodified.
 func Scrubber(args []string) []string {
-	badwords := []string{"username", "user", "password", "passwd", "secret"}
 	for _, arg := range args {
-		for _, bad := range badwords {
-			if strings.Contains(arg, bad) {
```
suneeth51: The original idea here was to check to perform either full string match or substring match.
For ex: If arg[0]= "password is here", then it must be filtered as bad (so Contains).

logger/logger.go
```diff
@@ -416,15 +416,28 @@ func HTTPLogger(inner http.Handler, name string) http.Handler {
 	})
 }
 
+// IsSensitive checks if the given key exists in the list of bad words (sensitive info)
+func IsSensitive(key string) bool {
```
suneeth51: Are we missing substring comparison?

NimbleMike: I'll update the PR to perform the substring comparison.

logger/logger.go
NimbleMike: Once I go back to the substring check, adding "accessSecret" will not be necessary.

PR #47, opened 2019-09-18
Moved windows specfic code for flexvol to common libs

vidhutsingh:


windows/mount.go
```diff
@@ -0,0 +1,61 @@
...
+	return nil
+}
+
+//GetDockerVolAccessPath - Get the docker volume access path of the partition
+func GetDockerVolAccessPath(flexvolPath string) (string, error) {
+	args := []string{"-command", "$vol = Get-Volume -FilePath", flexvolPath,
+		";", "(Get-Partition -Volume $vol).AccessPaths|where {$_ -match \".*hpe-storage-mounts.*\"}"}
+	dockerPath, rc, err := util.ExecCommandOutput("powershell", args)
+	if rc != 0 || err != nil {
+		util.LogError.Printf("Failed to get hpe-storage volume %s access path on the host, error %v", flexvolPath, err.Error())
```
raunakkumar: we have deprecated the the use of util.Logger, could you please switch to the new logger package else build will fail

vidhutsingh: thanks, I have replaced it with logger.done.

PR #51, opened 2019-10-01
chapi: add partition support for mounting imported volumes using csi …

shivamerla:
…plugin

* Problem:
  * partitions were not detected during import vol/clone requests and chapi
  * was attempting to recreate filesystem and reported failures during node stage.
* Implementation:
  * Following changes were made
  * Detect filesystem type on partition instead of whole device if partitions exist on device
  * Mount largest partition on device if one exists
  * Refactor mount logic to avoid duplicate attempts to mount whole device first
* Testing: tested import with partitions as well as regular volume mounts, unmount and cleanup.
* Review: gcostea, rkumar
* Bug: https://nimblejira.nimblestorage.com/browse/CON-491

### DIFF
--- a/linux/mount.go
+++ b/linux/mount.go
@@ -531,7 +531,7 @@ func MountDeviceWithFileSystem(devPath string, mountPoint string, options []stri
 	// If there are partitions on the device, then mount the largest partition.
 	deviceParitionInfos, err := GetPartitionInfo(&model.Device{AltFullPathName: devPath})
 	if err != nil {
-		return nil, fmt.Errorf("Failed to verify if partitions exists on device %s, %s", devPath, err.Error())
+		return nil, fmt.Errorf("Failed to verify if partitions exist on device %s, %s", devPath, err.Error())
 	}
 	if len(deviceParitionInfos) != 0 {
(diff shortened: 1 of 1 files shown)


linux/mount.go
```diff
@@ -550,36 +573,42 @@ func MountDeviceWithFileSystem(devPath string, mountPoint string, options []stri
...
 
 // GetFilesystemType returns the filesystem type if present else empty string
 func GetFilesystemType(devPath string) (string, error) {
 	log.Trace(">>>>> GetFilesystemType, devPath: ", devPath)
 	defer log.Trace("<<<<< GetFilesystemType")
 
+	// If there are partitions on the device, then return FS type of largest partition.
+	deviceParitionInfos, err := GetPartitionInfo(&model.Device{AltFullPathName: devPath})
+	if err != nil {
+		return "", fmt.Errorf("Failed to verify if partitions exists on device %s, %s", devPath, err.Error())
```
rgcostea: nit: if partitions exist

shivamerla: changed this..thanks.

PR #60, opened 2019-10-09
New branch

skshetti:


tunelinux/multipath.go
```diff
@@ -353,3 +355,19 @@ func ConfigureMultipath() (err error) {
...
+func ValidateMultipathconf() {
+	prefixes := []string{"defaults", "blacklist", "blacklist_exceptions", "devices", "device", "multipaths", "multipath"}
+	for _, prefix := range prefixes {
+		content, err := ioutil.ReadFile(linux.MultipathConf)
+		r := regexp.MustCompile(""+prefix+"\\n*\\s*{")
+		newcontents := r.ReplaceAll(content, []byte(""+prefix+" {"))
+		err = ioutil.WriteFile(linux.MultipathConf, []byte(newcontents), 0644)
+		
+		if err != nil {
+		panic(err)
```
raunakkumar: Please return error instead of panic

mpathconfig/configresource.go
```diff
@@ -237,15 +237,18 @@ func addOption(section *Section, option string) {
...
 
+//checks for the section
 func isSection(line string) bool {
-	line = strings.TrimSpace(line)
+	line = strings.TrimSpace(line)	
 	prefixes := []string{"defaults", "blacklist", "blacklist_exceptions", "devices", "device", "multipaths", "multipath"}
-	for _, prefix := range prefixes {
-		if strings.HasPrefix(line, prefix) {
+        for _, prefix := range prefixes {
+		r := regexp.MustCompile("^"+prefix+"\\s*{")
```
raunakkumar: Mustcompile could panic , rather use Compile with appropriate error handling

tunelinux/multipath.go
skshetti: I will make the changes.

mpathconfig/configresource.go
raunakkumar: We've tried to move to Compile from MustCompile for same reason for safety.

 If you want to use the same could you please move it to the const/var section so that it is compiled early? You can refer to linux/multipath.go or linux/device.go.

skshetti: would this be fine?
r, err := regexp.Compile(""^"+prefix+"\\s*{"")
if err != nil {
    log.Fatal(err)
}

skshetti: Sorry..I won't be using the log.Fatal(err)

raunakkumar: Yes.

mpathconfig/configresource.go
```diff
@@ -237,15 +237,21 @@ func addOption(section *Section, option string) {
...
-	line = strings.TrimSpace(line)
+//checks for the section
+func isSection(line string) (bool, error) {
+	line = strings.TrimSpace(line)	
 	prefixes := []string{"defaults", "blacklist", "blacklist_exceptions", "devices", "device", "multipaths", "multipath"}
-	for _, prefix := range prefixes {
-		if strings.HasPrefix(line, prefix) {
-			return true
+        for _, prefix := range prefixes {
+		r, err := regexp.Compile("^"+prefix+"\\s*{")
```
shivamerla: @skshetti i think we need to just make sure these keys are whole words. here are the possibilities for eg:

multipaths {
multipath
{
So, if we change to ```r, err := regexp.Compile("^"+prefix+"\\s+``` i think that should do.

tunelinux/multipath.go
```diff
@@ -353,3 +358,27 @@ func ConfigureMultipath() (err error) {
 	}
 	return nil
 }
+
+
+//remove newline between section name and '{'
```
shivamerla: With above comment, please check if this is necessary.

mpathconfig/configresource.go
skshetti: If we look for the keys as whole words what if there is something like this:
defaults {
             multipath         
}
i.e if there is a parameter multipath and no value..there isn't any check for the parameters without any values or wrong values except for the device section where the vendor is nimble.
So in this case "nimbletune --set  category multipath" would return error.

thats why i first modified the multipath.conf file by removing newlines between section name and { so that  "regexp.Compile("^"+prefix+"\\s*{")" would work fine in any case.

shivamerla: no, param names cannot be same as section names(they will never be). So if you compare whole worlds, thats good enough to parse sections. The customer issue was with param multipath_dir, where we were looking for multipath section.

skshetti: so param names shouldn't be similar to section names but there is no check being made for these parameters right? And the user can modify /etc/multipath.conf right? so what if he does this.?

shivamerla: multipath checks for them. If unknown params are added to multipath.conf, it will throw an error. you can try after adding them and run "multipath -ll". So, i think users adding unknown params is very remote. Just to minimize churn in this code(as its used by multiple plugins), we can handle this particular customer issue.

skshetti: Okay...so no need to remove the newlines between section name and { and just checking for the whole word would be enough right?

shivamerla: Yes, i think that should resolve the issue. Please try to test it out with various combinations(valid keys) and verify as well. Thanks!

mpathconfig/configresource.go
```diff
@@ -237,17 +237,33 @@ func addOption(section *Section, option string) {
...
-			return true
-		}
-	}
-	return false
+//checks for the section
+func isSection(line string) (bool, error) {
+        line = strings.TrimSpace(line)
+        prefixes := []string{"defaults", "blacklist", "blacklist_exceptions", "devices", "device", "multipaths", "multipath"}
+        for _, prefix := range prefixes {
+                r, err := regexp.Compile("^"+prefix+"\\s*{")
```
shivamerla: Please use this instead of two expressions: ```"^"+prefix+"\s*[{]"``` and add space for ``` err != nil ``` checks. linter didn't catch this?

mpathconfig/configresource.go
```diff
@@ -268,7 +284,11 @@ func ParseConfig(filePath string) (config *Configuration, err error) {
 	for scanner.Scan() {
 		line := scanner.Text()
 		if !(strings.HasPrefix(line, "#")) && len(line) > 0 {
-			if isSection(line) {
+			section_present, err := isSection(line)
+			if err!= nil {
```
shivamerla: same here, please add space between ``` err != nil ``` checks.

mpathconfig/configresource.go
skshetti: This expression ,"^"+prefix+"\\s*[{]"   won't work in the following case:
multipath
{

shivamerla: it does Shubha..https://goplay.space/#ER3D-iC4KoR

skshetti: for each line in multipath.conf  isSection(line) is called .. so isSection("multipath") will be called in one iteration and isSection("\n") in the next iteration...so it won't work....what you said is correct but here since we are calling the isSection with every line so it won't work here.

skshetti: isSection("{") in the next iteration*

shivamerla: section name has to be followed by `{` or else multipath will throw an error as below. We can look into automatically handling that separately as i believe good amount of testing is required.

```
devices
{
    device {
        path_selector        "service-time 0"
        fast_io_fail_tmo     5
        prio                 alua
        path_grouping_policy group_by_prio
        vendor               "Nimble"
        path_checker         tur
        rr_min_io_rq         1
        hardware_handler     "1 alua"
        product              "Server"
        no_path_retry        30
        rr_weight            uniform
        failback             immediate
        dev_loss_tmo         infinity
    }
}

# multipath -ll
Oct 15 17:15:29 | missing '{' on line 1 of /etc/multipath.conf
Oct 15 17:15:29 | invalid keyword '{' on line 2 of /etc/multipath.conf```

shivamerla:
@skshetti i am merging this, but in future please add "Signed-off-by: tag" as required by our DCO policy. You can set this in the default commit message.

PR #61, opened 2019-10-10
DW-109 - Fix , CHAPI default timeout is 1m for UnMount workflow which…

vidhutsingh:
Bug : DW-109 
Isssue : CHAPI client is getting timedout when multiple unmount are requested at the same time . 
Fix : The default timeout values in 1m which is less , increasing the default timeout to 5m . 


vidhutsingh:
> flexvolume dockerclient already has 5 minute timeout , does that not suffice ? If this is changed it impacts the dockerclient timeout too

Sorry didn't understand your comment, this is related to docker unmount workflow and it is failing constantly on Windows . As I can see the default timeout for chapclient is 5m for Mount wokflow however we have 1m timeout for Unmount workflow which is less and, causing unmount workflow issues on Windows.

raunakkumar:
The same unmount workflow is triggered from flexvolume too if both have 5 minutes timeout it would always timeout if this takes up the entire 5 minutes .
Can you debug further why the unmount takes so long.? We have throttling in unmount workflow controlled by defaultChannelCapacity so we don't process more requests than that.

Some pointers could be duplicate mountID from previous requests which weren't cleaned up because of which mountConflictDelay gets triggered in container provider

vidhutsingh:
> The same unmount workflow is triggered from flexvolume too if both have 5 minutes timeout it would always timeout if this takes up the entire 5 minutes .
> Can you debug further why the unmount takes so long.? We have throttling in unmount workflow controlled by defaultChannelCapacity so we don't process more requests than that.
> 
> Some pointers could be duplicate mountID from previous requests which weren't cleaned up because of which mountConflictDelay gets triggered in container provider

The defaultChannelCapacity is 30 where here we are running unmount of 10 mountpoints . So ,chanellCapacity is not an issue here also there is no stale mount point as all are unique and run first time. 
How flexvolume is impacted here since even if we add 5m timeout to CHAPI in unmount workflow?, the total timeout still would be 5m so it doesn't matter. Here we are adding timeout to CHAPI call only  whch is internally invoked during docker unmount workflow. 
I think this change would not have any adverse impact even to linux workflow.

raunakkumar:
10 unmounts take more than a minute ? Are the requests blocking somewhere ? Unmounts should be fairly quick
Also, Since the code is common for Linux and Windows the timeout impacts both. We have tested ~100 unmounts with current values and didn't face issues as throttling with channels helped. Changing this could impact our scale testing.

vidhutsingh:
> 10 unmounts take more than a minute ? Are the requests blocking somewhere ? Unmounts should be fairly quick
> Also, Since the code is common for Linux and Windows the timeout impacts both. We have tested ~100 unmounts with current values and didn't face issues as throttling with channels helped. Changing this could impact our scale testing.

Please don't compare this with Linux as on Windows things are different. Also this is just a timeout value which we are increasing for CHAPI, if on Linux it CHAPI responds fast it will not have any impact anyway.

vidhutsingh:
Sorry, accidentaly closed this pull request. Reopened.

vidhutsingh:
Change the code so that it doesn't have any effect on linux plugin .

raunakkumar:
Thanks

PR #62, opened 2019-10-10
chapi: Add ping test before iscsi login to avoid timeout errors

shivamerla:
* Problem:
  * If the target is not reachable from some interfaces, then iscsi login will
  * take long time 120s before timeout. This causes long delay during device
  * discovery
* Implementation:
  * Add simple ping test before iSCSI login. Currently this is added only if iscsi interfaces
  * are defined(cloud).
* Testing: Tested mount with multiple volumes on ec2 instance

### DIFF
--- a/linux/iscsi.go
+++ b/linux/iscsi.go
@@ -9,6 +9,7 @@ import (
 	"github.com/hpe-storage/common-host-libs/util"
 	ping "github.com/sparrc/go-ping"
 	"io/ioutil"
+	"net"
 	"os"
 	"regexp"
 	"strconv"
@@ -49,6 +50,7 @@ const (
 	PingCount               = 5
 	PingInterval            = 10 * time.Millisecond
 	PingTimeout             = 5 * time.Second
+	DefaultIscsiPort        = 3260
 )
 
 //type of Scope (volume, group)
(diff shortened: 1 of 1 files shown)


shivamerla:
> I think this works but it does mean that ICMP must be enabled. We've hit issues in some of our java code using isReachable. isReachable does the following: "Best effort is made by the implementation to try to reach the host, but firewalls and server configuration may block requests resulting in a unreachable status while some specific ports may be accessible. A typical implementation will use ICMP ECHO REQUESTs if the privilege can be obtained, otherwise it will try to establish a TCP connection on port 7 (Echo) of the destination host."
> 
> Because that's not always reliable, we switched to an attempt to connect on a socket to a given IP:Port combination. Our port of choice has been 80 since it's the most widely used TCP port. I think this approach would be equivalent to doing net.Dial("tcp", "IP:Port") in go. You might end up with more consistent results going that route.

Thanks George. I agree, if ICMP is blocked then we might skip all iSCSI logins. Shall i use the logic to attempt TCP connection to port 3260? This port is  used by iscsid for all iSCSI discovery and login, so that should be accessible.

rgcostea:
And port 3260 must be available for iSCSI anyway.  Great!

shivamerla:
> And port 3260 must be available for iSCSI anyway. Great!

Thanks. Done..

PR #63, opened 2019-10-14
Added custom timeout value for mountConfictDealy as it is causing doc…

vidhutsingh:
Problem : DW-67, Mounting the volume into multiple containers fail. 
Fix : 
The mountConflictDealy default time out value is 2 min which coincides with  docker CLI timeout value. In this scenario when the volume is already mounted to one container and, the user tries to mount the same volume to second container, the plugin mount workflows always goes in to processMountConflictDealy and wait for 2mins (default hardcoded timeout value) which results into docker cli (mount workflow ) almost always getting timed out with the following errors 
_

"C:\Program Files\Docker\docker.exe: Error response from daemon: error while mounting volume 'C:\ProgramData\hpe-storage-mounts\testvol16server\': Post http://localhost:8080/VolumeDriver.Mount: context deadline exceeded"

_  

There is no option to increase the timeout value for docker CLI. So to fix this I have introduced "MOUNT_CONFLICT_DELAY" which is optional. If the user sets this env, the default mountconflictdelay value is overwritten. 

### DIFF
--- a/dockerplugin/handler/mount.go
+++ b/dockerplugin/handler/mount.go
@@ -6,18 +6,19 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
-	"github.com/hpe-storage/common-host-libs/chapi"
-	"github.com/hpe-storage/common-host-libs/connectivity"
-	"github.com/hpe-storage/common-host-libs/dockerplugin/plugin"
-	"github.com/hpe-storage/common-host-libs/dockerplugin/provider"
-	log "github.com/hpe-storage/common-host-libs/logger"
-	"github.com/hpe-storage/common-host-libs/model"
 	"net/http"
 	"os"
 	"reflect"
 	"strconv"
 	"strings"
 	"time"
+
+	"github.com/hpe-storage/common-host-libs/chapi"
+	"github.com/hpe-storage/common-host-libs/connectivity"
+	"github.com/hpe-storage/common-host-libs/dockerplugin/plugin"
+	"github.com/hpe-storage/common-host-libs/dockerplugin/provider"
+	log "github.com/hpe-storage/common-host-libs/logger"
+	"github.com/hpe-storage/common-host-libs/model"
 )
 
 const (
@@ -100,7 +101,7 @@ func VolumeDriverMount(w http.ResponseWriter, r *http.Request) {
 	}
 
 	//2. this method does poll to container provider to check if other hosts are attached until mountConflictDelay
-	processMountConflictDelay(pluginReq.Name, providerClient, pluginReq, plugin.MountConflictDelay)
+	processMountConflictDelay(pluginReq.Name, providerClient, pluginReq, plugin.GetMountConflictDealy())
 
 	mapMutex.Lock(pluginReq.Name)
 	log.Debugf("taken lock for volume %s in Mount", pluginReq.Name)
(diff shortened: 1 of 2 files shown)


shivamerla:
@vidhutsingh looks like source of the issue to be waiting for mount conflict timeout even when the volume is mounted on same host? what is the reason for that? currently logic is to check InUse flag and also verify iSCSI/FC connections to make sure its current host. These checks might be failing for Windows host. We should find that cause.

vidhutsingh:
> @vidhutsingh looks like source of the issue to be waiting for mount conflict timeout even when the volume is mounted on same host? what is the reason for that? currently logic is to check InUse flag and also verify iSCSI/FC connections to make sure its current host. These checks might be failing for Windows host. We should find that cause.

The container provider is throwing error when nimbleGetVolumeInfo() is called during process mount conflict dealy: 

> time="2019-10-13T23:39:57-07:00" level=debug msg="About to decode the error response &{0x6730b0 {} 0x681470} into destination interface" file="client.go:231"
> time="2019-10-13T23:39:57-07:00" level=trace msg="empty body EOF" file="client.go:235"

Since the processMountConflictDealy() doesn't receive volume info, it waits for default timeout 2 min. This is the root of the problem. I checked the CP logs as well and there are no errors/exceptions around this time. 

The processMountConflict dealy doesn't handle a scenario where already mounted volume to one host1 is requested to be mounted to host2 . It will always go into default timeout 2m. Since processMountConflict dealy returns only when volume.InUse is false or there are logged in iscsi/fc session from the host1.

It would be ok to introduce new env variable to control processMountConflict() dealy behavior anyways this is harmless.

rgcostea:
So isn't the problem the fact that we do not decode the response from the container provider correctly?  I don't think we need to introduce another environment variable to work around that do we?  Shouldn't we instead debug that issue and find out what happened?

shivamerla:
@vidhutsingh before we enter processing conflicts, we fetch the current details. If it fails we return from there. But if it was successful, and GET is failing once in wait loop, we continue hoping we will be able to recover. So, we need to see why EOF is being returned for GET requests later on. Share us the logs, will take a look.

```	volume, err := nimbleGetVolumeInfo(containerProviderClient, pluginReq)
	// Error from nimbleGetVolumeInfo(), we should bail
	if err != nil {
		log.Tracef("unable to get volume information for %s. err=%s", volName, err.Error())
		return
	}```

shivamerla:
@vidhutsingh Just checked code, looks like we are not passing reference to responseError, hence decode() is not successful. Can you fix this and see actual error returned from CP?

vidhutsingh:
I checked the code and we don't return an error if decode fails with error "empty body EOF" . Instead, it returns nil. I added more traces to understand what went wrong as there were no errors returned by the Container Provider(I used responseError as well).  

With NOS version (5.1.3.0-661953-opt,Array IP : 10.18.27.200), this issue is consistently reproducible. 
I checked the volume info :

> ="<<<<< nimbleGetVolumeInfo" file="get.go:145"
> time="2019-10-15T00:32:59-07:00" level=trace msg="NimbleGet &{ID: Name:multi Size:0 Description: InUse:true Published:false BaseSnapID: ParentVolID: Clone:false Config:map[] Metadata:[] SerialNumber:c9f40d4ae86ba1106c9ce900a471b474 AccessProtocol: Iqn: DiscoveryIP: MountPoint: Status:map[] Chap:<nil> Networks:[] ConnectionMode: LunID: TargetScope: IscsiSessions:[0xc000381200 0xc000381220] FcSessions:[]}" file="mount.go:549"

As you can see the InUse = true , we don't return in this condition from processMountConflictDealy(), further, I checked isCurrentHostAttached in the code. volume.IscsiSessions doesn't have any InitiatorName and InitiatorIP . So CP didnt return these value in Volume.IscsiSession, ideally it should have since mount request is from same host . I dont know Why CP didnt return ? thus processMountConflictDealy exhausted the 2m timeout. 

 .

PR #65, opened 2019-10-18
Fixed IscsiSession and FCSession tag for initiators and wwpn

vidhutsingh:

processMountConflictDealy is always timeouts since the IscsiSession and FCsession uses incorrect tag .

### DIFF
--- a/model/types.go
+++ b/model/types.go
@@ -193,13 +193,13 @@ type Volume struct {
 
 // FcSession info
 type FcSession struct {
-	InitiatorWwpn string `json:"initiatorWwpn,omitempty"`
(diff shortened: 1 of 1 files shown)


shivamerla:
Thanks @vidhutsingh  and @rgcostea i remember Raunak had tested these, wondering if it got changed in container-provider code between Fiij.8 and Goa releases. However i have checked both branches of CP and seems like it never would have worked. @raunakkumar can you confirm?

raunakkumar:
I remember that i was able to compare and match iscsi and fc sessions on the host and from CP. Not sure how it would without this fix. Need to dig into what changed since then. Thanks @vidhutsingh @rgcostea

PR #69, opened 2019-10-24
DW-140 

vidhutsingh:
Problem : 
The dockerplugind thread crashes in cleanStaleMount() call . 

Fix : 
respMount is a double pointer which can be a valid pointer with empty array. Handled the condition properly.

### DIFF
--- a/dockerplugin/handler/unmount.go
+++ b/dockerplugin/handler/unmount.go
@@ -5,12 +5,14 @@ package handler
 import (
 	"encoding/json"
 	"errors"
+	"net/http"
+	"strings"
+
 	"github.com/hpe-storage/common-host-libs/chapi"
 	"github.com/hpe-storage/common-host-libs/connectivity"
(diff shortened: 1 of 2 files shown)


dockerplugin/handler/unmount.go
```diff
@@ -123,9 +125,14 @@ func VolumeDriverUnmount(w http.ResponseWriter, r *http.Request) {
...
 		if err != nil {
-			dr = DriverResponse{Err: err.Error()}
-			json.NewEncoder(w).Encode(dr)
-			return
+			if !strings.Contains(err.Error(), "object was not found on the system") {
+				dr = DriverResponse{Err: err.Error()}
+				json.NewEncoder(w).Encode(dr)
+				return
+			} else {
+				log.Errorln("Delete device called failed, chapi returned ", err.Error())
```
shivamerla: nit: call failed..

PR #74, opened 2019-10-28
dockerplugin: fix issue with reporting of iscsi/fc session info between fiji/goa

shivamerla:
* Problem:
  * Fiji rest-sdk reported iscsi/fc sessions using camelCase, while goa one reported
  * with snake_case, thus unmarhalling fails from dockerplugin.
* Implementation:
  * Add legacy attributes for iscsi/fc sessions to handle both versions.
* Testing:
* Review:gcostea, vsingh

### DIFF
--- a/model/types.go
+++ b/model/types.go
@@ -191,15 +191,32 @@ type Volume struct {
 	FcSessions     []*FcSession           `json:"fc_sessions,omitempty"`
 }
 
+// Workaround NOS 5.0.x vs 5.1.x responses with different case
 // FcSession info
 type FcSession struct {
-	InitiatorWwpn string `json:"initiator_wwpn,omitempty"`
+	InitiatorWwpn       string `json:"initiator_wwpn,omitempty"`
+	InitiatorWwpnLegacy string `json:"initiatorWwpn,omitempty"`
 }
 
 // IscsiSession info
 type IscsiSession struct {
-	InitiatorName string `json:"initiator_name,omitempty"`
-	InitiatorIP   string `json:"initiator_ip_addr,omitempty"`
+	InitiatorName       string `json:"initiator_name,omitempty"`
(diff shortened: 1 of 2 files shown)


raunakkumar:
Aah finally, I was testing with a Fiji Array back then. Thanks @skrishna

shivamerla:
@raunakkumar thanks to @rgcostea and @vidhutsingh for figuring out this issue..

model/types.go
```diff
@@ -191,15 +191,32 @@ type Volume struct {
...
 // IscsiSession info
 type IscsiSession struct {
-	InitiatorName string `json:"initiator_name,omitempty"`
-	InitiatorIP   string `json:"initiator_ip_addr,omitempty"`
+	InitiatorName       string `json:"initiator_name,omitempty"`
+	InitiatorNameLegacy string `json:"initiatorName,omitempty"`
+	InitiatorIP         string `json:"initiator_ip_addr,omitempty"`
+}
+
+func (s FcSession) InitiatorWwpnStr() string {
```
rgcostea: Can you drop the Str?

model/types.go
```diff
@@ -191,15 +191,32 @@ type Volume struct {
...
+	if s.InitiatorWwpnLegacy != "" {
+		return s.InitiatorWwpnLegacy
+	}
+	return s.InitiatorWwpn
+}
+
+func (s IscsiSession) InitiatorNameStr() string {
```
rgcostea: And here?

model/types.go
shivamerla: @rgcostea attribute name and function names will clash. Hence added Str. Let me know if we can call it something else?

rgcostea: wow.  i miss overloading.  I guess this name will have to do :)

shivamerla: :-)

PR #76, opened 2019-11-03
Perform new login attempt whenever auth-token is empty

suneeth51:


storageprovider/csp/container_storage_provider.go
```diff
@@ -131,6 +130,22 @@ func (provider *ContainerStorageProvider) login() (int, error) {
 // Currently, it will login again if the server responds with a status code of unauthorized.
 func (provider *ContainerStorageProvider) invoke(request *connectivity.Request) (int, error) {
 	request.Header = make(map[string]string)
+
+	// Perform login attempt when AuthToken is empty
+	if provider.AuthToken == "" {
+		log.Info("Cached auth-token is empty, attempting login to CSP")
```
shivamerla: can we log backend details as well for which login is being done or when failed?

PR #79, opened 2019-11-15
Avoid checking /dev/disk/by-path entries for VST targets

shivamerla:
Don't check for /dev/disk/by-path entries for VST targets and always
attempt login. Since the re-login attempt is more common with GST, we can
use those entries for lookup. This will avoid stale entry problem with
large number of targets repeatedly created/deleted.

### DIFF
--- a/linux/iscsi.go
+++ b/linux/iscsi.go
@@ -86,26 +86,27 @@ func RescanAndLoginToTarget(volume *model.Volume) (err error) {
 		return fmt.Errorf("Unable to retrieve Iscsi bound ifaces. Error: %s", err.Error())
 	}
 
-	loggedInTargets, err := GetIscsiTargets()
-	if err != nil {
-		log.Errorf("Unable to retrieve Iscsi Targets Error: %s", err.Error())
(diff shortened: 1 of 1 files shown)


linux/iscsi.go
```diff
@@ -86,26 +86,27 @@ func RescanAndLoginToTarget(volume *model.Volume) (err error) {
...
-	log.Trace("Logged in Targets :")
-	for _, target := range loggedInTargets {
-		log.Trace(target.Name)
-		if target.Name == volume.Iqn {
-			// best effort to scan new paths only with targetScope=GROUP
-			if strings.EqualFold(volume.TargetScope, GroupScope.String()) {
+	if strings.EqualFold(volume.TargetScope, GroupScope.String()) {
+		loggedInTargets, err := GetIscsiTargets()
+		if err != nil {
+			log.Errorf("Unable to retrieve Iscsi Targets Error: %s", err.Error())
```
rgcostea: nit: iSCSI

shivamerla: fixed it..

Issue #77, opened 2019-11-15
incorrect host network reported when secondary IP is present on an interface


knackaron:
When a node has a secondary IP on an interface, the driver reports the wrong network for the host.  This affects our master nodes that might have a secondary IP assigned via keepalived for the API server VIP in an on-premise cluster.

This could be worked around if we were able to define the discovery IP, as being discussed here: https://github.com/hpe-storage/csi-driver/issues/50

```
[root@k8s01b ~]# ip a show dev eth0
2: eth0: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc mq state UP group default qlen 1000
    link/ether 56:6f:ac:17:00:75 brd ff:ff:ff:ff:ff:ff
    inet 192.168.133.74/26 brd 192.168.133.127 scope global noprefixroute eth0
       valid_lft forever preferred_lft forever
    inet 192.168.133.70/32 scope global eth0
       valid_lft forever preferred_lft forever
    inet6 fe80::546f:acff:fe17:75/64 scope link noprefixroute
       valid_lft forever preferred_lft forever

[root@k8s01b ~]# kubectl -n kube-system get hpenodeinfos -oyaml k8s01b.test.example.com                                                                                                             
apiVersion: storage.hpe.com/v1
kind: HPENodeInfo
metadata:
  creationTimestamp: "2019-11-07T21:40:02Z"
  generation: 1
  name: k8s01b.test.example.com
  resourceVersion: "5193"
  selfLink: /apis/storage.hpe.com/v1/hpenodeinfos/k8s01b.test.example.com
  uid: 27dc6a42-01a7-11ea-a4c0-566fac170006
spec:
  iqns:
  - iqn.1994-05.com.redhat:1c149c6d5933
  networks:
  - 192.168.133.70
  - 172.16.16.0
  - 10.244.0.0
  uuid: 02428b1a-4316-6b38-7330-31622e746573
```

Versus a node without the secondary IP:
```
[root@k8s01b ~]# kubectl -n kube-system get hpenodeinfos -oyaml k8s02b.example.com
apiVersion: storage.hpe.com/v1
kind: HPENodeInfo
metadata:
  creationTimestamp: "2019-11-07T21:40:02Z"
  generation: 1
  name: k8s02b.example.com
  resourceVersion: "5181"
  selfLink: /apis/storage.hpe.com/v1/hpenodeinfos/k8s02b.example.com
[...]

PR #85, opened 2019-11-25
chapi : support for cidr networks

raunakkumar:
* Problem:
- support for cidr networks
- also support for multiple ip addresses on same interface https://github.com/hpe-storage/common-host-libs/issues/77
- support multiple discovery ips
* Implementation:
to minimize the impact to existing chapi,flexvolume introduce new fields for CSI specific
* Testing:
tested with multiple interfaces and interface with multiple ips
* Bug: https://nimblejira.nimblestorage.com/browse/CON-568

### DIFF
--- a/dockerplugin/handler/handler.go
+++ b/dockerplugin/handler/handler.go
@@ -136,14 +136,14 @@ type VersionResponse struct {
 
 // Host is a duplicate of model.Host which references a Network defined in this package
 type Host struct {
-	UUID           string                    `json:"id,omitempty"`
-	Name           string                    `json:"name,omitempty"`
-	Domain         string                    `json:"domain,omitempty"`
-	NodeID         string                    `json:"node_id,omitempty"`
-	AccessProtocol string                    `json:"access_protocol,omitempty"`
-	Networks       []*model.NetworkInterface `json:"networks,omitempty"`
-	Initiators     []*model.Initiator        `json:"initiators,omitempty"`
-	Version        string                    `json:"version,omitempty"`
+	UUID              string                    `json:"id,omitempty"`
+	Name              string                    `json:"name,omitempty"`
+	Domain            string                    `json:"domain,omitempty"`
+	NodeID            string                    `json:"node_id,omitempty"`
+	AccessProtocol    string                    `json:"access_protocol,omitempty"`
+	NetworkInterfaces []*model.NetworkInterface `json:"networks,omitempty"`
+	Initiators        []*model.Initiator        `json:"initiators,omitempty"`
+	Version           string                    `json:"version,omitempty"`
 }
 
 // populate hostcontext into request and fetch user info if running in cloud vm
--- a/dockerplugin/handler/handler.go
+++ b/dockerplugin/handler/handler.go
@@ -136,14 +136,14 @@ type VersionResponse struct {
 
 // Host is a duplicate of model.Host which references a Network defined in this package
 type Host struct {
-	UUID           string             `json:"id,omitempty"`
-	Name           string             `json:"name,omitempty"`
-	Domain         string             `json:"domain,omitempty"`
-	NodeID         string             `json:"node_id,omitempty"`
-	AccessProtocol string             `json:"access_protocol,omitempty"`
-	Networks       []*model.Network   `json:"networks,omitempty"`
-	Initiators     []*model.Initiator `json:"initiators,omitempty"`
-	Version        string             `json:"version,omitempty"`
+	UUID           string                    `json:"id,omitempty"`
+	Name           string                    `json:"name,omitempty"`
+	Domain         string                    `json:"domain,omitempty"`
+	NodeID         string                    `json:"node_id,omitempty"`
+	AccessProtocol string                    `json:"access_protocol,omitempty"`
+	Networks       []*model.NetworkInterface `json:"networks,omitempty"`
+	Initiators     []*model.Initiator        `json:"initiators,omitempty"`
+	Version        string                    `json:"version,omitempty"`
 }
 
 // populate hostcontext into request and fetch user info if running in cloud vm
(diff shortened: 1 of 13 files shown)


linux/network.go
```diff
@@ -110,14 +110,46 @@ func GetIPV4NetworkAddress(ipv4Address, netMask string) (networkAddress string,
...
-func GetNetworkInterfaces() ([]*model.Network, error) {
-	log.Trace(">>>>> GetNetworkInterfaces")
+func GetNetworkInterfaces(cidrOnly bool) ([]*model.Network, error) {
+	log.Tracef(">>>>> GetNetworkInterfaces called with %v", cidrOnly)
 	defer log.Trace("<<<<< GetNetworkInterfaces")
 
+	if cidrOnly {
+		networks, err := getCIDRNetworks()
+		return networks, err
+	}
```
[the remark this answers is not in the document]
raunakkumar: make sense will make the change

linux/device.go
```diff
@@ -870,7 +870,7 @@ func CreateNimbleDevices(vols []*model.Volume) (devs []*model.Device, err error)
 	var devices []*model.Device
 	for _, vol := range vols {
 		log.Tracef("create request with serialnumber :%s, accessprotocol %s discoveryip %s, iqn %s ", vol.SerialNumber, vol.AccessProtocol, vol.DiscoveryIP, vol.Iqn)
-		if vol.AccessProtocol == iscsi && (vol.DiscoveryIP == "" || vol.Iqn == "") {
+		if vol.AccessProtocol == iscsi && (vol.DiscoveryIP == "" || vol.Iqn == "") && len(vol.DiscoveryIPs) == 0 {
```
shivamerla: i think this should be || instead of &&, as older versions doesn't set DiscoveryIP's, but just DiscoveryIP

linux/iscsi.go
```diff
@@ -501,6 +513,18 @@ func validateChapUserPassword(user []string, password []string) (chapInfo *model
 	chapInfo = &model.ChapInfo{Name: user[0], Password: password[0]}
 	return chapInfo, nil
 }
+func PerformDiscoveries(discoveryIPs []string) (a model.IscsiTargets, err error) {
```
shivamerla: we should not do this. We need to check any one discovery IP which is reachable. thats good enough, it gives targets from all portals and we can login.

linux/iscsi.go
```diff
@@ -112,11 +112,23 @@ func RescanAndLoginToTarget(volume *model.Volume) (err error) {
...
-	targetSet, err := PerformDiscovery(volume.DiscoveryIP)
-	if err != nil {
-		log.Errorf("Unable to Perform Discovery with discoveryIp: %s. Error: %s", volume.DiscoveryIP, err.Error())
-		return fmt.Errorf("Unable to Perform Discovery with discoveryIp: %s. Error: %s", volume.DiscoveryIP, err.Error())
+	var targetSet model.IscsiTargets
+	if len(volume.DiscoveryIPs) == 0 && volume.DiscoveryIP == "" {
+		return fmt.Errorf("no discovery ip present to discover iSCSI Targets")
+	}
+	// give preference to slice of discovery IPs over a single discovery IP
+	if len(volume.DiscoveryIPs) > 0 {
```
shivamerla: i think we  need to change PerformDiscovery() to just take an array, and find any one reachable IP from it and use that for discovery.

linux/device.go
raunakkumar: since this is error condition check which means neither of discoveryIp or discoveryIps are set that is why && condition if any of them is set we should be fine.

linux/iscsi.go
raunakkumar: sure will refactor this. any corner cases to consider?

dockerplugin/handler/handler.go
```diff
@@ -136,14 +136,14 @@ type VersionResponse struct {
...
-	AccessProtocol string             `json:"access_protocol,omitempty"`
-	Networks       []*model.Network   `json:"networks,omitempty"`
-	Initiators     []*model.Initiator `json:"initiators,omitempty"`
-	Version        string             `json:"version,omitempty"`
+	UUID           string                    `json:"id,omitempty"`
+	Name           string                    `json:"name,omitempty"`
+	Domain         string                    `json:"domain,omitempty"`
+	NodeID         string                    `json:"node_id,omitempty"`
+	AccessProtocol string                    `json:"access_protocol,omitempty"`
+	Networks       []*model.NetworkInterface `json:"networks,omitempty"`
```
rgcostea: Can we also rename Networks to NetworkInterfaces

linux/iscsi.go
```diff
@@ -504,19 +518,40 @@ func validateChapUserPassword(user []string, password []string) (chapInfo *model
...
+		if isDiscoveryIpReachable {
+			args := []string{"-m", "discovery", "-t", "st", "-p", discoveryIP, "-o", "new"}
+			out, _, err = util.ExecCommandOutput(iscsicmd, args)
+			if err != nil {
+				log.Error(err.Error())
+				continue
+			}
+			break
+		}
+	}
```
rgcostea: Nice!

model/types.go
```diff
@@ -280,7 +282,7 @@ type FcHostPort struct {
 // Iface represents iface configuring with port binding
 type Iface struct {
 	Name    string
-	Network *Network
+	Network *NetworkInterface
```
rgcostea: And here.

model/types.go
```diff
@@ -180,10 +181,11 @@ type Volume struct {
 	AccessProtocol string                 `json:"access_protocol,omitempty"`
 	Iqn            string                 `json:"iqn,omitempty"`
 	DiscoveryIP    string                 `json:"discovery_ip,omitempty"` // this field needs to be moved out ?
+	DiscoveryIPs   []string               `json:"discovery_ips,omitempty"`
 	MountPoint     string                 `json:"Mountpoint,omitempty"`
 	Status         map[string]interface{} `json:"status,omitempty"` // interface so that we can map any number of arguments
 	Chap           *ChapInfo              `json:"chap_info,omitempty"`
-	Networks       []*Network             `json:"networks,omitempty"`
+	Networks       []*NetworkInterface    `json:"networks,omitempty"`
```
rgcostea: And here.

PR #87, opened 2019-11-26
Setting replacement timeout for iscsi sessions

skshetti:


tunelinux/iscsi.go
```diff
@@ -364,6 +364,11 @@ func SetIscsiRecommendations() (err error) {
 					continue
 				}
 			}
+			if recommendation.Parameter == "replacement_timeout" {
```
shivamerla: @skshetti i thought we will introduce --global parameter to configure settings at host level or bypass otherwise. Are you not taking that approach?

tunelinux/iscsi.go
```diff
@@ -417,3 +422,41 @@ func ConfigureIscsi() (err error) {
 
 	return nil
 }
+
+// Update replacement timeout for logged-in sessions
+func UpdateIscsiSessionReplacementTimeout() (err error) {
```
shivamerla: I think we can use generic function to update this like any other parameter. We don't need separate one for each parameter to apply.

shivamerla:
@skshetti  please add Signed-off-by: tag when next commits.

tunelinux/iscsi.go
```diff
@@ -364,11 +364,14 @@ func SetIscsiRecommendations() (err error) {
 					continue
 				}
 			}
-			// Modify iscsid.conf settings accordingly
-			err = SetIscsiParamRecommendation(recommendation.Parameter, recommendation.Recommendation)
-			if err != nil {
-				// continue with other recommendations
-				continue
+
+			if !global {
```
shivamerla: This should be other way

``` if global {```

File: tunelinux/iscsi.go [hunk 4]
```diff
@@ -389,6 +392,7 @@ func SetIscsiRecommendations() (err error) {
 
 // ConfigureIscsi verifies and install necessary iSCSI packages if not present. It also makes sure if service is running.
 func ConfigureIscsi() (err error) {
+	global := false
```
shivamerla: make global as argument here too..i will have to change all callers(across dcs/cloudvol, csi-driver etc) after this to pass this boolean as true(as most of the times we need host level setting)

skshetti: ConfigureIscsi is being called from dockervolume component. There are multiple levels involved in the calling of ConfigureIscsi and for host level settings nimbletune --set category -global will take care right?

PR #89, opened 2019-12-04
storageprovider: misc fixes

raunakkumar:
* Problem:
- allow login from one backend to happen only once
- Http error 200 gets logged during recursive invocation
of login
* Implementation:
- have mutex per backend for login
- check for status code 200 and return immediately, have single variable
for status and error so that the values are not scope bound
* Testing:
* Bug: https://nimblejira.nimblestorage.com/browse/CON-573

### DIFF
--- a/storageprovider/csp/container_storage_provider.go
+++ b/storageprovider/csp/container_storage_provider.go
@@ -16,6 +16,7 @@ import (
 	"google.golang.org/grpc/codes"
 	grpcstatus "google.golang.org/grpc/status"
 
+	"github.com/hpe-storage/common-host-libs/concurrent"
 	"github.com/hpe-storage/common-host-libs/connectivity"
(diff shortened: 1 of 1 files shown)


storageprovider/csp/container_storage_provider.go
```diff
@@ -484,6 +495,11 @@ func (provider *ContainerStorageProvider) GetVolume(id string) (*model.Volume, e
 			ResponseError: &errorResponse,
 		},
 	)
+
+	if status == http.StatusOK {
+		return dataWrapper.Data.(*model.Volume), nil
+	}
+
```
shivamerla: Since we are returning here on success, the end statement of function should be `return nil, err`?

raunakkumar: yup, ideally we should never reach there but will modify the end statement

PR #91, opened 2019-12-05
Somewhat duplicate PR, this time with test

e4jet:
The main thing to look at here is  linux/mount_test.go

A new test exercised the *WithOptions functions.  It take 18 seconds to run, but wont run unless mkfs.xfs is installed (which it isn't in most containers).

### DIFF
--- a/model/types.go
+++ b/model/types.go
@@ -2,6 +2,8 @@ package model
 
 import (
 	"fmt"
+	"regexp"
+	"strings"
 )
 
 const (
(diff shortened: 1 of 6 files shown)


model/types.go
```diff
@@ -298,9 +300,23 @@ type Hosts []*Host
...
+	Type       string
+	Mode       string
+	Owner      string
+	CreateOpts string
+}
+
+// GetCreateOpts returns a clean array that can be passed to the command line
+func (f FilesystemOpts) GetCreateOpts() []string {
+	cleanCharRe := regexp.MustCompile(`[^a-zA-Z0-9=, \-]`)
+	singleSpacesRe := regexp.MustCompile(`\s+`)
```
rgcostea: Maybe it's just me but it's not obvious that Re on the end means Regex.  Can we just make this cleanCharRegex?

e4jet: Sure.  Code change here: https://github.com/hpe-storage/common-host-libs/pull/92

PR #93, opened 2019-12-11
chapi: fix misc issues and support other vendors

shivamerla:
* Problem:
  * serial_number provided to chapi is assumed to be with scsi-id type prefix 2 (mpath)
  * iscsi targets are looked up by nimble specific patterns and device filtering used nimble vendor types
  * dependency on /dev/disk/by-path symlinks for logged-in sessions, but group wide targets might be logged-in
  * without devices
* Implementation:
  * change to always assume serial number input to chapi is of volume serial number and same with returned types
  * only consider scsi-id type prefix during mpath comparisions, not store anywhere
  * add table for vendor patterns for device and target-iqn comparision and filtering
  * fix function names to be generic
* Testing: in-progress with CSI/FlexVolume drivers

### DIFF
--- a/chapi/chapiclient.go
+++ b/chapi/chapiclient.go
@@ -430,7 +430,7 @@ func (chapiClient *Client) GetDeviceFromVolume(volume *model.Volume) (device *mo
 	if err != nil {
 		return nil, err
 	}
-	serialNumber := GetSerialNumber(volume.SerialNumber)
+	serialNumber := volume.SerialNumber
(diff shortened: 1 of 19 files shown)


shivamerla:
Tested pod create/delete with CSI and verified serial number passed correctly and iSCSI settings are applied.

shivamerla:
@raunakkumar updated path uuid too without prefix and tested with flexvolume as well. @wdurairaj  please pickup latest code and test it out with 3PAR.

PR #95, opened 2019-12-11
serialize access of /proc/mounts

raunakkumar:
* Problem:
- during nodeUnpublish several threads
try to access /proc/mounts inside the container to retrive
the bind mounts. We've seen that there are inconsistent outputs
even for existing mounts
* Implementation:
- take exclusive mutex lock for accessing /proc/mounts
* Testing:
* Bug: https://nimblejira.nimblestorage.com/browse/CON-597

### DIFF
--- a/linux/bmount.go
+++ b/linux/bmount.go
@@ -18,6 +18,7 @@ package linux
 
 import (
 	"fmt"
+	"github.com/hpe-storage/common-host-libs/concurrent"
 	log "github.com/hpe-storage/common-host-libs/logger"
 	"github.com/hpe-storage/common-host-libs/util"
 	"strings"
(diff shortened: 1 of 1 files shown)


linux/bmount.go
```diff
@@ -25,9 +25,10 @@ import (
 )
 
 const (
-	procMounts    = "/proc/mounts"
-	mountCommand  = "mount"
-	umountCommand = "umount"
+	procMounts     = "/proc/mounts"
+	mountCommand   = "mount"
+	umountCommand  = "umount"
+	procMountMutex = concurrent.NewMapMutex()
```
shivamerla: it should be a variable instead of const.

raunakkumar: done

PR #98, opened 2020-01-30
chapi: avoid /proc/mounts due to duplicates

shivamerla:
* Problem:
  * When using chroot mechanism, both /var/lib/kubelet and /host will be mounted within container
  * This will cause duplicate mount entries in /proc/mounts leading confusion in mount cleanup.
  * Also, dmsetup remove is hanging within chroot.
* Implementation:
  * Use "mount" call to fetch mount points rather than /proc/mounts as its inconsistent. mount call always
  * yeilds without duplicates.
  * use multipathd remove instead of dmsetup remove to avoid hang issue.
* Testing: regression tests on RHEL 7.6, and pending on Ubuntu 16 and 18.
* Review: gcostea, rkumar

### DIFF
--- a/chapi/chapidriver_linux.go
+++ b/chapi/chapidriver_linux.go
@@ -290,7 +290,7 @@ func (driver *LinuxDriver) BindMount(mountPoint string, newMountPoint string, rb
 // BindUnmount unmounts the given bind mount
 func (driver *LinuxDriver) BindUnmount(mountPoint string) error {
 	log.Tracef(">>>>> BindUnmount, mountPoint: %s", mountPoint)
-	defer log.Trace("<<<<< BindMount")
+	defer log.Trace("<<<<< BindUnmount")
 
 	// Unmount given bind mount
 	return linux.RetryBindUnmount(mountPoint)
--- a/linux/bmount.go
+++ b/linux/bmount.go
@@ -18,21 +18,20 @@ package linux
 
 import (
 	"fmt"
-	"github.com/hpe-storage/common-host-libs/concurrent"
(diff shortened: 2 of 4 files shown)


shivamerla:
```
Example of duplicates when host root is mounted within container as /host.

/dev/mapper/mpathbt /var/lib/kubelet/plugins/hpe.com/mounts/0634be4e62e74eae4d00000000000000000000009a xfs rw,relatime,attr2,inode64,noquota 0 0
/dev/mapper/mpathbt /var/lib/kubelet/pods/10944406-42ef-11ea-ad82-0050569633fd/volumes/kubernetes.io~csi/pvc-f75b5bfb-42ee-11ea-ad82-0050569633fd/mount xfs rw,relatime,attr2,inode64,noquota 0 0

/dev/mapper/mpathbt /host/var/lib/kubelet/plugins/hpe.com/mounts/0634be4e62e74eae4d00000000000000000000009a xfs rw,relatime,attr2,inode64,noquota 0 0
/dev/mapper/mpathbt /host/var/lib/kubelet/pods/10944406-42ef-11ea-ad82-0050569633fd/volumes/kubernetes.io~csi/pvc-f75b5bfb-42ee-11ea-ad82-0050569633fd/mount xfs rw,relatime,attr2,inode64,noquota 0 0```

raunakkumar:
Lgtm. We should hold off on merge until we have few automation runs on csi and other consumers.

shivamerla:
> Lgtm. We should hold off on merge until we have few automation runs on csi and other consumers.

Yes, agreed.

PR #102, opened 2020-03-13
multipathd reload doesn't update size of the map, use resize instead and error out on timeout errors.

shivamerla:

### DIFF
--- a/linux/device.go
+++ b/linux/device.go
@@ -1081,10 +1081,13 @@ func RescanForCapacityUpdates(devicePath string) error {
 		// multipathd takes either mpathx(without prefix) or /dev/dm-x as input
 		devicePath = strings.TrimPrefix(devicePath, "/dev/mapper/")
 		// reload multipath map to apply new size
-		args = []string{"reload", "map", devicePath}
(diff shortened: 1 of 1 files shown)


linux/device.go
```diff
@@ -1081,11 +1081,14 @@ func RescanForCapacityUpdates(devicePath string) error {
 		// multipathd takes either mpathx(without prefix) or /dev/dm-x as input
 		devicePath = strings.TrimPrefix(devicePath, "/dev/mapper/")
 		// reload multipath map to apply new size
-		args = []string{"reload", "map", devicePath}
-		_, _, err = util.ExecCommandOutput("multipathd", args)
+		args = []string{"resize", "map", devicePath}
```
raunakkumar: Didn't we attempt resize 5 times prior to reaching here? 
if multipathd was timing out before, any reason it wouldn't timeout now. Is the fix to just attempt another try?

shivamerla: Yes, exactly. We need to let NodeExpand to retry as if multipathd keeps timing out we cannot update the device size.

PR #117, opened 2020-05-08
chapi: add nfs mount endpoint

shivamerla:
* Problem:
  * need capability in chapi to perform nfs mounts
* Implementation:
  * add endpoint which performs just mount without device checking.
* Testing: tested with nfs volumes using csi driver.
* Review: gcostea, rkumar

### DIFF
--- a/linux/mount.go
+++ b/linux/mount.go
@@ -570,6 +570,28 @@ func MountDeviceWithFileSystem(devPath string, mountPoint string, options []stri
 	return mount, nil
 }
 
+func MountNFSShare(source string, targetPath string, options []string) error {
+	log.Tracef(">>>>> MountNFSShare called with source %s target %s", source, targetPath)
+	defer log.Tracef("<<<<< MountNFSShare")
+
+	args := []string{source, targetPath}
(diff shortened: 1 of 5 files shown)


linux/mount.go
```diff
@@ -570,6 +570,28 @@ func MountDeviceWithFileSystem(devPath string, mountPoint string, options []stri
...
 }
 
+func MountNFSShare(source string, targetPath string, options []string) error {
+	log.Tracef(">>>>> MountNFSShare called with source %s target %s", source, targetPath)
+	defer log.Tracef("<<<<< MountNFSShare")
+
+	args := []string{source, targetPath}
+	optionArgs := []string{}
+	if len(options) != 0 {
+		optionArgs = append([]string{"-o"}, strings.Join(options, ","))
```
raunakkumar, line 580: should we validate or restrict it to the ones we tested or want to keep it opaque?

shivamerla: yes, it should be opaque. we have nfsMountOptions parameter to tweak these. But we can document any options that doesn't work well.

PR #119, opened 2020-05-15
chapi: provide option to specify nfs version type

shivamerla:
* Problem:
  * Some flavors default to nfs4 and others to nfs4.1 causing mounts to fail if
  * server doesn't support 4.1
* Implementation:
  * Add nfsType option to mount
* Testing: tested with multiple nfs mounts using csi

### DIFF
--- a/chapi/chapidriver_linux.go
+++ b/chapi/chapidriver_linux.go
@@ -235,11 +235,11 @@ func (driver *LinuxDriver) GetMountsForDevice(device *model.Device) ([]*model.Mo
 	return linux.GetMountPointsForDevices(devices)
 }
 
-func (driver *LinuxDriver) MountNFSVolume(source string, targetPath string, mountOptions []string) error {
+func (driver *LinuxDriver) MountNFSVolume(source string, targetPath string, mountOptions []string, nfsType string) error {
 	log.Tracef(">>>>> MountNFSVolume called with source %s target %s options %v: ", source, targetPath, mountOptions)
 	defer log.Trace("<<<<< MountNFSVolume")
 
-	err := linux.MountNFSShare(source, targetPath, mountOptions)
+	err := linux.MountNFSShare(source, targetPath, mountOptions, nfsType)
 	if err != nil {
 		return fmt.Errorf("Error mounting nfs share %s at %s, err %s", source, targetPath, err.Error())
 	}
--- a/linux/mount.go
+++ b/linux/mount.go
@@ -570,11 +570,16 @@ func MountDeviceWithFileSystem(devPath string, mountPoint string, options []stri
 	return mount, nil
 }
 
-func MountNFSShare(source string, targetPath string, options []string) error {
+func MountNFSShare(source string, targetPath string, options []string, nfsType string) error {
 	log.Tracef(">>>>> MountNFSShare called with source %s target %s", source, targetPath)
 	defer log.Tracef("<<<<< MountNFSShare")
 
-	args := []string{source, targetPath}
+	// default nfs version to 4.0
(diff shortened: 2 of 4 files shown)


chapi/chapidriver_linux.go
```diff
@@ -235,11 +235,11 @@ func (driver *LinuxDriver) GetMountsForDevice(device *model.Device) ([]*model.Mo
 	return linux.GetMountPointsForDevices(devices)
 }
 
-func (driver *LinuxDriver) MountNFSVolume(source string, targetPath string, mountOptions []string) error {
+func (driver *LinuxDriver) MountNFSVolume(source string, targetPath string, mountOptions []string, nfsType string) error {
 	log.Tracef(">>>>> MountNFSVolume called with source %s target %s options %v: ", source, targetPath, mountOptions)
```
suneeth51, line 239: Update trace log for `nfsType` as well

linux/mount.go
```diff
@@ -570,11 +570,16 @@ func MountDeviceWithFileSystem(devPath string, mountPoint string, options []stri
...
 }
 
-func MountNFSShare(source string, targetPath string, options []string) error {
+func MountNFSShare(source string, targetPath string, options []string, nfsType string) error {
 	log.Tracef(">>>>> MountNFSShare called with source %s target %s", source, targetPath)
 	defer log.Tracef("<<<<< MountNFSShare")
 
-	args := []string{source, targetPath}
+	// default nfs version to 4.0
+	if nfsType == "" {
```
raunakkumar, line 578: nit: name this nfsVersion?

linux/mount.go
```diff
@@ -570,11 +570,16 @@ func MountDeviceWithFileSystem(devPath string, mountPoint string, options []stri
 	return mount, nil
 }
 
-func MountNFSShare(source string, targetPath string, options []string) error {
+func MountNFSShare(source string, targetPath string, options []string, nfsType string) error {
 	log.Tracef(">>>>> MountNFSShare called with source %s target %s", source, targetPath)
```
suneeth51, line 574: same comment as above to include nfsType in log

linux/mount.go
```diff
@@ -570,11 +570,16 @@ func MountDeviceWithFileSystem(devPath string, mountPoint string, options []stri
...
 	defer log.Tracef("<<<<< MountNFSShare")
 
-	args := []string{source, targetPath}
+	// default nfs version to 4.0
+	if nfsType == "" {
+		nfsType = "nfs4"
```
suneeth51, line 579: Move this hard-coded string to constants.h

linux/mount.go
shivamerla: Changed comment to reflect as nfsType instead. As its referred as filesystem type in general.

```For  most types all the mount program has to do is issue a simple mount(2) system call, and no detailed knowledge of the filesystem type
              is required.  For a few types however (like nfs, nfs4, cifs, smbfs, ncpfs) an ad hoc code is necessary.  The nfs, nfs4, cifs, smbfs, and
              ncpfs  filesystems  have a separate mount program.  In order to make it possible to treat all types in a uniform way, mount will execute
              the program /sbin/mount.type (if that exists) when called with type type.```

chapi/chapidriver_linux.go
shivamerla: added..thanks

linux/mount.go
shivamerla: done..

linux/mount.go
raunakkumar: cool

PR #123, opened 2020-05-22
snapshotgroup: support for create/delete snapshot group

raunakkumar:

### DIFF
--- a/model/types.go
+++ b/model/types.go
@@ -252,6 +252,16 @@ type VolumeGroup struct {
 	Config       map[string]interface{} `json:"config,omitempty"`
 }
 
+type SnapshotGroup struct {
+	ID                    string                 `json:"id,omitempty"`
+	Name                  string                 `json:"name,omitempty"`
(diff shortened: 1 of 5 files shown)


model/types.go
```diff
@@ -252,6 +252,16 @@ type VolumeGroup struct {
 	Config       map[string]interface{} `json:"config,omitempty"`
 }
 
+type SnapshotGroup struct {
+	ID                    string                 `json:"id,omitempty"`
+	Name                  string                 `json:"name,omitempty"`
+	SourceVolumeGroupID   string                 `json:"volume_group_id,omitempty"`
+	SourceVolumeGroupName string                 `json:"volume_group_name,omitempty"`
```
rgcostea, line 259: Do we need to put Source as a prefix here?

raunakkumar: CSP still treats it without source. Should i change across or keep it as is.?
```    
@JsonProperty("volume_group_id")
@JsonProperty("volume_group_name")
```

PR #126, opened 2020-06-29
Accomodate scanning for multiple lun id

wdurairaj:
- Current in 3PAR/Primera CSP we are adding requirement for Peer persistence based Replication, where we need to create host and vlun's for both the Primary (source) array and the Target (secondary) array, so that the multipath device mapping will have paths from active/passive arrays. But, the publish volume response from CSP to CSI has currently support for only returning a single lun id. The change here is to return a object called SecondaryArrayDetails which will have list of object PeerDetails which has lun id, discovery ip, target names etc. One PeerDetail will be created per secondary backend.

- Approach

Changes are done to find, if the multiple luns are passed based on the number of PeerDetails in SecondaryArrayDetails struct. iscsi login , scanning all takes care of reading this object, and invokes appropriate actions.

Signed-off-by: William Durairaj <w_durairaj@yahoo.com>

### DIFF
--- a/model/types.go
+++ b/model/types.go
@@ -166,36 +166,37 @@ type DevicePartition struct {
 
 // Volume : Thin version of Volume object for Host side
 type Volume struct {
-	ID                   string                 `json:"id,omitempty"`
-	Name                 string                 `json:"name,omitempty"`
-	Size                 int64                  `json:"size,omitempty"` // size in bytes
-	Description          string                 `json:"description,omitempty"`
-	InUse                bool                   `json:"in_use,omitempty"` // deprecated for published in the CSP implementation
-	Published            bool                   `json:"published,omitempty"`
-	BaseSnapID           string                 `json:"base_snapshot_id,omitempty"`
-	ParentVolID          string                 `json:"parent_volume_id,omitempty"`
-	Clone                bool                   `json:"clone,omitempty"`
-	Config               map[string]interface{} `json:"config,omitempty"`
-	Metadata             []*KeyValue            `json:"metadata,omitempty"`
-	SerialNumber         string                 `json:"serial_number,omitempty"`
-	AccessProtocol       string                 `json:"access_protocol,omitempty"`
-	Iqn                  string                 `json:"iqn,omitempty"` // deprecated
-	Iqns                 []string               `json:"iqns,omitempty"`
-	DiscoveryIP          string                 `json:"discovery_ip,omitempty"` // deprecated
-	DiscoveryIPs         []string               `json:"discovery_ips,omitempty"`
-	MountPoint           string                 `json:"Mountpoint,omitempty"`
-	Status               map[string]interface{} `json:"status,omitempty"` // interface so that we can map any number of arguments
-	Chap                 *ChapInfo              `json:"chap_info,omitempty"`
-	Networks             []*NetworkInterface    `json:"networks,omitempty"`
-	ConnectionMode       string                 `json:"connection_mode,omitempty"`
-	LunID                string                 `json:"lun_id,omitempty"`
-	SecondaryLunIDs      string                 `json:"secondary_lun_ids,omitempty"`
-	SecondaryTargetNames string                 `json:"secondary_target_names,omitempty"`
-	SecondaryDiscoverIps string                 `json:"secondary_discovery_ips,omitempty"`
-	TargetScope          string                 `json:"target_scope,omitempty"` //GST="group", VST="volume" or empty(older array fiji etc), and no-op for FC
-	IscsiSessions        []*IscsiSession        `json:"iscsi_sessions,omitempty"`
-	FcSessions           []*FcSession           `json:"fc_sessions,omitempty"`
-	VolumeGroupId        string                 `json:"volume_group_id"`
+	ID                    string                 `json:"id,omitempty"`
+	Name                  string                 `json:"name,omitempty"`
+	Size                  int64                  `json:"size,omitempty"` // size in bytes
+	Description           string                 `json:"description,omitempty"`
+	InUse                 bool                   `json:"in_use,omitempty"` // deprecated for published in the CSP implementation
+	Published             bool                   `json:"published,omitempty"`
+	BaseSnapID            string                 `json:"base_snapshot_id,omitempty"`
+	ParentVolID           string                 `json:"parent_volume_id,omitempty"`
+	Clone                 bool                   `json:"clone,omitempty"`
+	Config                map[string]interface{} `json:"config,omitempty"`
+	Metadata              []*KeyValue            `json:"metadata,omitempty"`
+	SerialNumber          string                 `json:"serial_number,omitempty"`
+	AccessProtocol        string                 `json:"access_protocol,omitempty"`
+	Iqn                   string                 `json:"iqn,omitempty"` // deprecated
+	Iqns                  []string               `json:"iqns,omitempty"`
+	DiscoveryIP           string                 `json:"discovery_ip,omitempty"` // deprecated
+	DiscoveryIPs          []string               `json:"discovery_ips,omitempty"`
+	MountPoint            string                 `json:"Mountpoint,omitempty"`
+	Status                map[string]interface{} `json:"status,omitempty"` // interface so that we can map any number of arguments
+	Chap                  *ChapInfo              `json:"chap_info,omitempty"`
+	Networks              []*NetworkInterface    `json:"networks,omitempty"`
+	ConnectionMode        string                 `json:"connection_mode,omitempty"`
+	LunID                 string                 `json:"lun_id,omitempty"`
+	SecondaryLunIDs       string                 `json:"secondary_lun_ids,omitempty"`
+	SecondaryTargetNames  string                 `json:"secondary_target_names,omitempty"`
+	SecondaryDiscoverIps  string                 `json:"secondary_discovery_ips,omitempty"`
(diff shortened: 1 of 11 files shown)


model/types.go
```diff
@@ -284,7 +284,7 @@ type AccessInfo struct {
 type BlockDeviceAccessInfo struct {
 	AccessProtocol string   `json:"access_protocol,omitempty"`
 	TargetNames    []string `json:"target_names,omitempty"`
-	LunID          int32    `json:"lun_id,omitempty"`
+	LunID          string   `json:"lun_id,omitempty"`
```
rgcostea, line 287: I think the type should directly state what the expectation is.  In this case, it should be []int32.

rgcostea, line 287: This change needs to make its way into the CSP spec definition here: https://github.com/hpe-storage/container-storage-provider/blob/master/spec.md.

wdurairaj: Fixed in the recent update.

wdurairaj: Raised PR -- https://github.com/hpe-storage/container-storage-provider/pull/14 for this spec change.

model/types.go
```diff
@@ -285,6 +286,19 @@ type BlockDeviceAccessInfo struct {
 	AccessProtocol string   `json:"access_protocol,omitempty"`
 	TargetNames    []string `json:"target_names,omitempty"`
 	LunID          int32    `json:"lun_id,omitempty"`
+	SecondaryBackendDetails
+	IscsiAccessInfo
+}
+
+// Information of LUN id, IQN, discovery IP's the secondary array
+type SecondaryBackendDetails struct {
+	PeerArrayDetails []SecondaryLunInfo
```
rgcostea, line 295: We typically do a []*SecondaryLunInfo.  Can you see above?

util/volume.go
```diff
@@ -0,0 +1,71 @@
+package util
```
rgcostea, line 1: missing copyright header

util/volume.go
```diff
@@ -0,0 +1,71 @@
...
+
+	var volObj *model.Volume
+	volObj.SerialNumber = serialNumber
+	volObj.LunID = lunID
+
+	return volObj
+}
+
+func GetSecondaryArrayLUNIds(details string) []int32 {
+	var secondaryArrayDetails model.SecondaryBackendDetails
```
rgcostea, line 19: We have this logger pattern that we use in many of our functions...
>>>>> function name
<<<<< defer function name

to make debugging easier.  Can you do the same here?

util/volume.go
```diff
@@ -0,0 +1,71 @@
...
+		logger.Tracef("\n Error in GetSecondaryArrayLUNIds %s", err.Error())
+		logger.Tracef("\n Passed string: %s", details)
+		return []int32{}
+	}
+	numberOfSecondaryBackends := len(secondaryArrayDetails.PeerArrayDetails)
+	var secondaryLunIds []int32 = make([]int32, numberOfSecondaryBackends)
+	for i := 0; i < numberOfSecondaryBackends; i++ {
+		secondaryLunIds[i] = secondaryArrayDetails.PeerArrayDetails[i].LunID
+	}
+	return secondaryLunIds
```
rgcostea, line 32: Won't this return all secondary LUN IDs on all secondary arrays?  I thought we wanted to handle each secondaryArray independent of any other secondaryArray.  Isn't that more correct?

util/volume.go
```diff
@@ -0,0 +1,71 @@
...
+		return []string{}
+	}
+	numberOfSecondaryBackends := len(secondaryArrayDetails.PeerArrayDetails)
+	var secondaryTargetNames []string
+	for i := 0; i < numberOfSecondaryBackends; i++ {
+		for _, targetNameRetrieved := range secondaryArrayDetails.PeerArrayDetails[i].TargetNames {
+			secondaryTargetNames = append(secondaryTargetNames, targetNameRetrieved)
+		}
+	}
+	return secondaryTargetNames
```
rgcostea, line 51: Same comment here

util/volume.go
```diff
@@ -0,0 +1,71 @@
...
+		logger.Tracef("\n Passed details %s", details)
+		return []string{}
+	}
+	numberOfSecondaryBackends := len(secondaryArrayDetails.PeerArrayDetails)
+	var secondaryDiscoverIps []string
+	for i := 0; i < numberOfSecondaryBackends; i++ {
+		for _, discoveryIpRetrieved := range secondaryArrayDetails.PeerArrayDetails[i].DiscoveryIPs {
+			secondaryDiscoverIps = append(secondaryDiscoverIps, discoveryIpRetrieved)
+		}
+	}
```
rgcostea, line 69: And here.

util/volume.go
wdurairaj: Yes it will return all the secondary targets lun ids. While do a rescanFc/IscsiTarget() what we need is the list of lun id to iterate and issue "echo - - <lun-id>" >  /sys/... " so processing each lun id from individual secondary target or clubbing together all the lun id's (as done here) , the rescan logic is done for all luns together. what i have done is the seggregation in the logic for the Primary lun scan and all the luns for individual secondary targets.

model/types.go
wdurairaj: i will incorporate this

util/volume.go
wdurairaj: i will incorporate this

util/volume.go
wdurairaj: i will incorporate this

util/volume.go
rgcostea: Will let @shivamerla comment here.  Our approach was to have a high level function to iterate over the secondary array details and perform the login + scan logic on each of them.  Clubbing all of these IDs together seems to defeat the purpose of having stored them in this fashion.

shivamerla: @wdurairaj lets discuss this over call. I think what we want to achieve is a way to validate if all secondary luns are discovered and claimed by multipath. Lets see if we can add an additional call which does secondaryLunAttach() and perform discovery/scan on each. So, we don't touch existing logic anywhere and handle secondary ones separately.

wdurairaj: @rgcostea , @shivamerla , i have done some refactoring around handling secondary arrays, however the handleRemapLun(), handleOrphanPaths() i have kept the code intact, since i could'nt meaningfully split that code across primary and secondary backends(s) since once the volume's path are discovered, we need to handle the paths (based on lunids) as a single group only.

Let me know, if the change is looking ok or any other changes required.
I fixed the other review comments from @rgcostea  as well.

I have done testing around CSP response with/without secondary backends

util/volume.go
```diff
@@ -0,0 +1,93 @@
...
+
+func GetSecondaryBackends(details string) []*model.SecondaryLunInfo {
+	logger.Tracef(">>>>> GetSecondaryBackends %s", details)
+	defer logger.Tracef("<<<< GetSecondaryBackends")
+	var secondaryArrayDetails model.SecondaryBackendDetails
+	logger.Tracef("\n About to unmarshal %s", details)
+	err := json.Unmarshal([]byte(details), &secondaryArrayDetails)
+	if err != nil {
+		logger.Tracef("\n Error in GetSecondaryBackends %s", err.Error())
+		logger.Tracef("\n Passed details %s", details)
```
shivamerla, line 88: you can remove this additional trace as its already added as entry point.

util/volume.go
```diff
@@ -0,0 +1,93 @@
...
+}
+
+func GetSecondaryArrayDiscoveryIps(details string) []string {
+	logger.Tracef(">>>>> GetSecondaryArrayDiscoveryIps %s", details)
+	defer logger.Tracef("<<<< GetSecondaryArrayDiscoveryIps")
+	var secondaryArrayDetails model.SecondaryBackendDetails
+	logger.Tracef("\n About to unmarshal %s", details)
+	err := json.Unmarshal([]byte(details), &secondaryArrayDetails)
+	if err != nil {
+		logger.Tracef("\n Error in GetSecondaryArrayDiscoveryIps %s", err.Error())
```
shivamerla, line 67: please change this to Errorf()

util/volume.go
```diff
@@ -0,0 +1,93 @@
...
+}
+
+func GetSecondaryArrayTargetNames(details string) []string {
+	logger.Tracef(">>>>> GetSecondaryArrayTargetNames %s", details)
+	defer logger.Tracef("<<<< GetSecondaryArrayTargetNames")
+	var secondaryArrayDetails model.SecondaryBackendDetails
+	logger.Tracef("\n About to unmarshal %s", details)
+	err := json.Unmarshal([]byte(details), &secondaryArrayDetails)
+	if err != nil {
+		logger.Tracef("\n Error in GetSecondaryArrayTargetNames %s", err.Error())
```
shivamerla, line 47: same here..Errorf()

util/volume.go
```diff
@@ -0,0 +1,93 @@
...
+}
+
+func GetSecondaryArrayLUNIds(details string) []int32 {
+	logger.Tracef(">>>>> GetSecondaryArrayLUNIds %s", details)
+	defer logger.Tracef("<<<< GetSecondaryArrayLUNIds")
+	var secondaryArrayDetails model.SecondaryBackendDetails
+	logger.Tracef("\n About to unmarshal %s", details)
+	err := json.Unmarshal([]byte(details), &secondaryArrayDetails)
+	if err != nil {
+		logger.Tracef("\n Error in GetSecondaryArrayLUNIds %s", err.Error())
```
shivamerla, line 29: same here..

PR #129, opened 2020-07-05
Added support for TrueNAS Core CSP

datamattsson:
- Allow discovery of TrueNAS Core disks and targets.
- Added multipath.conf stanzas

### DIFF
--- a/linux/iscsi.go
+++ b/linux/iscsi.go
@@ -57,7 +57,7 @@ const (
 
 var (
 	iscsiMutex           sync.Mutex
-	targetVendorPatterns = []string{"com.nimblestorage", "com.3pardata"}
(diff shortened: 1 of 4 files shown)


linux/iscsi.go
```diff
@@ -57,7 +57,7 @@ const (
 
 var (
 	iscsiMutex           sync.Mutex
-	targetVendorPatterns = []string{"com.nimblestorage", "com.3pardata"}
+	targetVendorPatterns = []string{"com.nimblestorage", "com.3pardata", "org.truenas.ctl"}
```
shivamerla, line 60: We can look into removing these altogether. As filtering is not required and we can always rely on full target IQN name and serial provided during node stage.

PR #130, opened 2020-07-21
chapi: update chap credentials for all targets

raunakkumar:
- chap user and password were only updated for a single target/portal.
- update chap user/password for all the target/portals matching the targetIqn

### DIFF
--- a/linux/iscsi.go
+++ b/linux/iscsi.go
@@ -117,6 +117,39 @@ func getReachableDiscoveryPortals(discoveryIPs []string, virtualPortal bool) (re
 	return reachablePortals, nil
 }
 
+func updateChapForLoggedInTargets(volume *model.Volume) {
+	log.Tracef(">>>>> updateChapForLoggedInTargets for volume %s", volume.SerialNumber)
+	defer log.Tracef("<<<<< updateChapForLoggedInTargets")
+	loggedInTargets, _ := GetIscsiNodesFromIsciadm()
+	var chapUser, chapPassword string
+	if volume.Chap == nil {
+		chapUser = ""
+		chapPassword = ""
+	} else {
+		chapUser = volume.Chap.Name
+		chapPassword = volume.Chap.Password
+	}
+	for _, targetName := range volume.TargetNames() {
+		for _, loggedInTarget := range loggedInTargets {
+			// update only the targets matching the target iqn
+			if loggedInTarget.Name == targetName {
+				log.Tracef("updating chapUser for Targets :%s Portal :%s to %s", loggedInTarget.Name, loggedInTarget.Address, chapUser)
+				// update chapuser irrespective (could be a toggle case chap->empty)
+				err := updateChapUser(loggedInTarget, chapUser)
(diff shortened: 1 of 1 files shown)


linux/iscsi.go
```diff
@@ -220,11 +221,11 @@ func loginToTarget(targets model.IscsiTargets, targetIqn string, ifaces []*model
 			log.Debug("Found target :", "Target:", targetIqn)
 			// Update Chap credentials
 			if chapUser != "" && chapPassword != "" {
-				err = updateChapUser(target, chapUser)
+				err = updateChapUser(target, targets, chapUser)
```
shivamerla, line 224: Raunak, looks like we are looping through each target:portal pair for chap and connection method too. Can we just change addTarget() alone to login to each target:portal pair instead and not loop within each function?

raunakkumar: Sure. The addTarget method was already overloaded with/without ifaces. Also connection method looks to be updated after the login. Any repercussions of any of these if we preset it before login?

shivamerla: I think we just need to add portal here within addTarget()
```
args := []string{"--mode", "node", "--targetname", target.Name, "--login"}
```
and remove break after each login and don't error out if single target login fails. So any login fails we don't update connection method for that target:portal pair alone and we might update during next volume stage.

Also, here we need to consider CHAP being toggled. i.e if chap username/password is empty, but previously they were applied, we need to reset with each session too.

linux/iscsi.go
```diff
@@ -193,9 +233,8 @@ func HandleIscsiDiscovery(volume *model.Volume) (err error) {
 		return err
 	}
 
-	if !loggedIn {
```
shivamerla, line 196: looks like you can directly call updateChapFoLoggedInTargets() here in else case instead of passing to another function.

linux/iscsi.go
```diff
@@ -300,43 +322,33 @@ func isReachable(initiatorIP, targetIP string) (reachable bool, err error) {
...
 	return
 }
 
 // updates iscsi node db with given chap username
 func updateChapUser(target *model.IscsiTarget, chapUser string) (err error) {
-	log.Tracef(">>>>> updateChapUser for target %s user %s", target.Name, chapUser)
+	log.Tracef(">>>>> updateChapUser for target %s portal %s, user %s", target.Name, target.Address, chapUser)
 	defer log.Trace("<<<<< updateChapUser")
 
-	iscsiMutex.Lock()
```
shivamerla, line 337: lets keep these, they will prevent iscsi db corruption.

raunakkumar: they cause nested lock exception since addTarget takes the lock before this

linux/iscsi.go
raunakkumar: sure

linux/iscsi.go
shivamerla: ok..got it..iscsi mutex is held there..

PR #134, opened 2020-08-28
common-host-libs: Regex to detect orphan paths using single digit SCSI host number in H:C:T:L format

shivamerla:
* Problem:
  * Due to current regex, only devices with single digit host numbers in (H:C:T:L) are
  * filtered to detect orphan paths. This is causing delay in device attach when
  * paths are kept in orphan state by multipathd.
* Implementation:
  * Fix the regex to consider space before H:C:T:L numbers, so SCSI host numbers like
  * 10, 11 etc are detected as well.
* Testing: Tested on scale setup to ensure orphan paths are detected correctly to speed up discovery.
* Review: gcostea, sbyadarahalli, rkumar

### DIFF
--- a/linux/multipath.go
+++ b/linux/multipath.go
@@ -27,7 +27,7 @@ const (
 	MultipathConf = "/etc/multipath.conf"
 	// MultipathBindings bindings file for multipathd
 	MultipathBindings  = "/etc/multipath/bindings"
-	orphanPathsPattern = ".*(?P<host>\\d+):(?P<channel>\\d+):(?P<target>\\d+):(?P<lun>\\d+).*(REPLACE_VENDOR).*orphan"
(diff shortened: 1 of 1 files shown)


shivamerla:
Output with the issue:

time="2020-08-27T18:03:16Z" level=debug msg="orphan h:c:t:l found is **0**:0:0:65" file="multipath.go:426"
time="2020-08-27T18:03:16Z" level=debug msg="orphan h:c:t:l found is **1**:0:0:65" file="multipath.go:426"
time="2020-08-27T18:03:16Z" level=debug msg="orphan h:c:t:l found is **2**:0:0:65" file="multipath.go:426"
time="2020-08-27T18:03:16Z" level=debug msg="orphan h:c:t:l found is **3**:0:0:65" file="multipath.go:426"

PR #136, opened 2020-08-31
Collect volume metrics to be sent to array-exporter and CSI

ckallianpur:


storageprovider/csp/container_storage_provider.go
```diff
@@ -691,6 +691,37 @@ func (provider *ContainerStorageProvider) GetVolumeByName(name string) (*model.V
 	return volume, err
 }
 
+// GetVolumeMetrics will return stats about the given volume
+func (provider *ContainerStorageProvider) GetVolumeMetrics(id string) (*model.VolumeMetrics, error) {
```
rgcostea, line 695: Why do you think we need a separate function for this?  We already call the same endpoint for GetVolume which is where we will add these new fields.  Furthermore, I think the VolumeMetrics struct above is unnecessary.  We should just add the new fields we need.  Per the CSI spec they are:

>   // The available capacity in specified Unit. This field is OPTIONAL.
  // The value of this field MUST NOT be negative.
  int64 available = 1;
  // The total capacity in specified Unit. This field is REQUIRED.
  // The value of this field MUST NOT be negative.
  int64 total = 2;
  // The used capacity in specified Unit. This field is OPTIONAL.
  // The value of this field MUST NOT be negative.
  int64 used = 3;

We should simply report total and used space and then infer the available space from these 2.

ckallianpur: I will simplify this.

PR #138, opened 2020-11-12
Modified files for handling multiple iscsi session logout

bhagyashree-sarawate:
Problem statement:
There was an issue with iscsiadm logout step as part of the cleanup of deletedevices in NodeUnstageVolume. It deletes and logs out only 1 target node.
As part of nodeStageVolume, CSI creates deviceInfo.json and writes staging device information in it. But, it writes only one iscsiTarget info in it rather than a list

The mentioned PR has the fix for the multiple iscsi session logout.
It now writes the list of the iscsiTargets in deviceInfo.json file so that it logs out all the iscsi session.

### DIFF
--- a/linux/iscsi.go
+++ b/linux/iscsi.go
@@ -819,11 +819,12 @@ func removeDuplicateTargets(targets model.IscsiTargets) model.IscsiTargets {
 }
 
 // iscsiGetTargetOfDevice : get the iscsi target information of a device from sysfs
-func iscsiGetTargetOfDevice(dev *model.Device) (target *model.IscsiTarget, err error) {
+func iscsiGetTargetOfDevice(dev *model.Device) (target []*model.IscsiTarget, err error) {
 	log.Trace(">>>>> iscsiGetTargetOfDevice  with", dev)
 	defer log.Trace("<<<<< iscsiGetTargetOfDevice")
 
 	r := regexp.MustCompile(sessionIDPattern)
+	iscsiTargets := make([]*model.IscsiTarget, 0)
 	for _, hcil := range dev.Hcils {
 		host := strings.Split(hcil, ":")[0]
 		iscsiHostPath := fmt.Sprintf(hostDeviceFormat, host)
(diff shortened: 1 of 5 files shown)


model/types.go
```diff
@@ -150,11 +150,12 @@ type Device struct {
 	MpathName       string       `json:"mpath_device_name,omitempty"`
 	Size            int64        `json:"size,omitempty"` // size in MiB
 	Slaves          []string     `json:"slaves,omitempty"`
-	IscsiTarget     *IscsiTarget `json:"iscsi_target,omitempty"`
+	IscsiTarget     []*IscsiTarget `json:"iscsi_target,omitempty"`
```
raunakkumar, line 153: modify to 
```IscsiTargets     []*IscsiTarget `json:"iscsi_targets,omitempty"` ``` ?

File: linux/iscsi.go [hunk 2]
```diff
@@ -849,10 +850,16 @@ func iscsiGetTargetOfDevice(dev *model.Device) (target *model.IscsiTarget, err e
 			if err != nil {
 				return nil, fmt.Errorf("no iscsi target found from iscsi session %s", sessionID)
 			}
-			return iscsiTarget, nil
+			iscsiTargets = append(iscsiTargets, iscsiTarget)
 		}
 	}
-	return nil, nil
+	if iscsiTargets != nil {
```
raunakkumar, line 856: just return `iscsiTargets, nil` instead of checking the if condition

linux/iscsi.go
```diff
@@ -819,11 +819,12 @@ func removeDuplicateTargets(targets model.IscsiTargets) model.IscsiTargets {
 }
 
 // iscsiGetTargetOfDevice : get the iscsi target information of a device from sysfs
-func iscsiGetTargetOfDevice(dev *model.Device) (target *model.IscsiTarget, err error) {
+func iscsiGetTargetOfDevice(dev *model.Device) (target []*model.IscsiTarget, err error) {
```
raunakkumar, line 822: rename this method to `iscsiGetTargetsOfDevice`

linux/iscsi.go
raunakkumar: The caller of this method would also need to be modified. 
Currently we set the device iscsiTarget but now we need to refer to iscsiTargets[0] atleast for non 3Pardata
```iscsiTarget, err := iscsiGetTargetOfDevice(device)
				if err != nil {
					log.Debugf("unable to get iscsi target for device %s. Error: %s", device.Pathname, err.Error())
					continue
					// continue with other devices, might be in offline state and IscsiTarget will be nil for this device
				}
				if iscsiTarget != nil {
					device.IscsiTarget = iscsiTarget
				}
			}
```

raunakkumar (changes requested):
Please make the suggested changes and could we have a private image to be run by nimble automation team to make sure there is no regression?

sneharai4 (approved):

wdurairaj (approved):
Changes looks fine for me.

raunakkumar (approved):

PR #139, opened 2020-12-03
CON-1537 : Skip color coding in windows

ckallianpur:


logger/logger.go
```diff
@@ -274,7 +274,10 @@ func (hook *ConsoleHook) Fire(entry *log.Entry) error {
 
 	// Write log entry to output stream
 	if logParams.UseTextFormatter() {
-		hook.formatter.(*log.TextFormatter).ForceColors = hook.checkIfTerminal(logWriter)
+		//https://github.com/sirupsen/logrus/issues/172
+		if runtime.GOOS != "windows" {
+			hook.formatter.(*log.TextFormatter).ForceColors = hook.checkIfTerminal(logWriter)
```
rkumpf, line 279: I don't understand why this statement is needed at all.  Logrus itself has similar logic:

	func (f *TextFormatter) isColored() bool {
		isColored := f.ForceColors || (f.isTerminal && (runtime.GOOS != "windows"))

		if f.EnvironmentOverrideColors {
			switch force, ok := os.LookupEnv("CLICOLOR_FORCE"); {
			case ok && force != "0":
				isColored = true
			case ok && force == "0", os.Getenv("CLICOLOR") == "0":
				isColored = false
			}
		}

		return isColored && !f.DisableColors
	}

It seems like either we don't trust the logrus terminal check or we want to force colors for Windows.

Is another project depending on the colorization for logging to a Windows terminal?  If not, I'd suggest removing the whole code block at lines 276-281.

rkumpf: I see now that common-host-libs disables the logrus logging, and only uses logrus for formatting.  So when logrus checks if it is writing to a terminal, the check will always be false.  The terminal check is needed here because common-host-libs is managing the output streams itself.

Based on all that, I suggest disabling colors on Windows like in your original patch.  I don't know if any existing project wants colors on a Windows terminal, but it seems like it wouldn't work reliably anyway.

rkumpf (approved):

raunakkumar (approved):

PR #150, opened 2021-05-18
chapi: generate unique uuid for 2 hosts with same mac address

raunakkumar:
Problem:
As seen in issue https://github.com/hpe-storage/csi-driver/issues/270
Two hosts have same mac addresses and although their hostnames differ,
we generate the same uuid as the mechanism to generate uuid does not guarantee
uniqueness.
Update the mechanism to pass it through md5 has to guarantee uniqueness for 32 characters

Signed-off-by: Raunak <raunak.kumar@hpe.com>

### DIFF
--- a/chapi/chapidriver_linux.go
+++ b/chapi/chapidriver_linux.go
@@ -72,7 +72,7 @@ func generateUniqueNodeId() (string, error) {
 	// sort mac addresses as best attempt to get same address even if node.gob is deleted
 	sort.Strings(macs)
 
-	// create a unique id using mac addess and host name using Md5 hash
+	// create a unique id using mac address and host name using Md5 hash
 	// https://github.com/hpe-storage/csi-driver/issues/270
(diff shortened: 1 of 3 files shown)


chapi/chapidriver_linux.go
```diff
@@ -73,8 +72,9 @@ func generateUniqueNodeId() (string, error) {
 	// sort mac addresses as best attempt to get same address even if node.gob is deleted
 	sort.Strings(macs)
 
-	// create a unique id using mac addess and host name
-	idStr := strings.Replace(macs[0], ":", "", -1) + hex.EncodeToString([]byte(hostName))
+	// create a unique id using mac addess and host name using Md5 hash
```
rgcostea, line 75: mac address

util/strings_test.go
```diff
@@ -0,0 +1,38 @@
+package util
```
rgcostea, line 1: copyright

sneharai4 (approved):

suneeth51 (approved):
lgtm..

PR #151, opened 2021-05-24
add 3par device section to multipath template

raunakkumar:
Signed-off-by: Raunak <raunak.kumar@hpe.com>

### DIFF
--- a/tunelinux/config/multipath.conf.generic
+++ b/tunelinux/config/multipath.conf.generic
@@ -62,4 +62,19 @@ devices {
         path_selector        "queue-length 0"
         hardware_handler     "1 alua"
     }
+    device {
+        product              “VV”
+        path_grouping_policy group_by_prio
+        features             “0"
+        prio                 alua
+        no_path_retry        18
+        hardware_handler     “1 alua”
+        failback             immediate
+        rr_min_io            100
+        vendor               “3PARdata”
+        getuid_callout       “/lib/udev/scsi_id --whitelisted --device=/dev/%n”
+        path_selector        “round-robin 0"
+        path_checker         tur
+        checker              tur
+    }
 }
--- a/tunelinux/config/multipath.conf.upstream
+++ b/tunelinux/config/multipath.conf.upstream
@@ -64,4 +64,19 @@ devices {
         path_selector        "queue-length 0"
         hardware_handler     "1 alua"
     }
+    device {
+        product              “VV”
+        path_grouping_policy group_by_prio
+        features             “0"
+        prio                 alua
+        no_path_retry        18
+        hardware_handler     “1 alua”
+        failback             immediate
+        rr_min_io            100
+        vendor               “3PARdata”
+        getuid_callout       “/lib/udev/scsi_id --whitelisted --device=/dev/%n”
+        path_selector        “round-robin 0"
+        path_checker         tur
+        checker              tur
+    }
 }


datamattsson:
@raunakkumar can you please check if these files actually gets copied to the node when running Ubuntu, we're seeing something else, only the Nimble stanza.

datamattsson:
This line here is moot:
https://github.com/hpe-storage/csi-driver/blob/9718cbf1b5c473292454a5f3ffdbc20ecea207dc/cmd/csi-driver/conform/entrypoint.sh#L37
Because of this:
https://github.com/hpe-storage/csi-driver/blob/9718cbf1b5c473292454a5f3ffdbc20ecea207dc/cmd/csi-driver/Dockerfile#L52

datamattsson:
I can’t make out why the wrong decisions are made on Ubuntu, it works as expected on CentOS.

raunakkumar:
> This line here is moot:
> https://github.com/hpe-storage/csi-driver/blob/9718cbf1b5c473292454a5f3ffdbc20ecea207dc/cmd/csi-driver/conform/entrypoint.sh#L37
> Because of this:
> https://github.com/hpe-storage/csi-driver/blob/9718cbf1b5c473292454a5f3ffdbc20ecea207dc/cmd/csi-driver/Dockerfile#L52

In the Dockerfile we pull the respective templates locally and in entrypoint.sh we copy the upstream template if it is missing on the node. /host of the container is mapped to the root of the node.

raunakkumar:
> I can’t make out why the wrong decisions are made on Ubuntu, it works as expected on CentOS.

As per Chris, the upstream Ubuntu does not have  3Par Multipath config whereas Centos/RHEL has it.

datamattsson:
Both Chris and the customer end up with a multipath.conf file that looks like this (on Ubuntu):
```
defaults {
    user_friendly_names yes
}
devices {
    device {
        features             0
        prio                 alua
        rr_min_io            100
        path_grouping_policy group_by_prio
        hardware_handler     "1 alua"
        failback             immediate
        path_selector        "round-robin 0"
        product              "Server"
        fast_io_fail_tmo     5
        path_checker         tur
        dev_loss_tmo         infinity
        rr_weight            uniform
        vendor               "Nimble"
        no_path_retry        18
    }
}
```
We need to understand why.

raunakkumar:
> Both Chris and the customer end up with a multipath.conf file that looks like this (on Ubuntu):
> 
> ```
> defaults {
>     user_friendly_names yes
> }
> devices {
>     device {
>         features             0
>         prio                 alua
>         rr_min_io            100
>         path_grouping_policy group_by_prio
>         hardware_handler     "1 alua"
>         failback             immediate
>         path_selector        "round-robin 0"
>         product              "Server"
>         fast_io_fail_tmo     5
>         path_checker         tur
>         dev_loss_tmo         infinity
>         rr_weight            uniform
>         vendor               "Nimble"
>         no_path_retry        18
>     }
> }
> ```
> 
> We need to understand why.
> 
> EDIT: The point I'm getting at is that this PR is moot for what we want to do unless there's any PRs outstanding in the other repos that address this.

@datamattsson  Just did an experiment on my Ubuntu Cluster with this PR

* Before Multipath.conf
```
ls -al /etc/multipath.conf
ls: cannot access '/etc/multipath.conf': No such file or directory
```
* After installing the CSI Driver with the updated template definitions 
```
root@hitdev-tc8:~# ls -al /etc/multipath.conf
-rw-r--r-- 1 root root 2286 May 24 14:03 /etc/multipath.conf

 cat /etc/multipath.conf | grep  -B 4 3PAR
        product "Server"
    }
    device {
        product "VV"
        vendor  "3PARdata"
--
    device {
        path_grouping_policy group_by_prio
        prio                 alua
        no_path_retry        18
        vendor               “3PARdata”

```

* The issue will still exist if there is an existing multipath.conf because as of our current device section validation, we only validate Nimble device section. To address this we need to work on enhancing multipath recommendations for 3par, but if the aim is to copy the correct template, this PR should be able to do the work.
https://github.com/hpe-storage/common-host-libs/blob/master/tunelinux/multipath.go#L21

datamattsson:
>     * The issue will still exist if there is an existing multipath.conf because  our current device section validations only validates Nimble device section.

So we're out on the deep end here. It's more than the device stanzas we need to care about. What is the reasoning behind that we're not using `/etc/multipath/conf.d/` and simply drop our configuration file in there to have the configuration merged with whatever the customer already has in place on their worker nodes? We obviously need to query `multipathd` what "config_dir" is set to:

```
# multipathd list config | grep config_dir | awk '{print $2}'
"/etc/multipath/conf.d"
```

raunakkumar:
> > ```
> > * The issue will still exist if there is an existing multipath.conf because  our current device section validations only validates Nimble device section.
> > ```
> 
> So we're out on the deep end here. It's more than the device stanzas we need to care about. What is the reasoning behind that we're not using `/etc/multipath/conf.d/` and simply drop our configuration file in there to have the configuration merged with whatever the customer already has in place on their worker nodes? We obviously need to query `multipathd` what "config_dir" is set to:
> 
> ```
> # multipathd list config | grep config_dir | awk '{print $2}'
> "/etc/multipath/conf.d"
> ```

Agree. Sure even if we dropped the file in the config_dir path we need to validate individual params to make sure it abides by recommended values. For us setting it to /etc/multipath.conf gives us less overhead to figure out what that config_dir could be.

datamattsson:
> For us setting it to /etc/multipath.conf gives us less overhead to figure out what that config_dir could be.

This will still create breakage on Ubuntu as we're seeing multiple customers running into "find_multipaths" being set to yes. How do we add the rest of the parameters we need to an existing multipath.conf file?

raunakkumar:
We validate different sections of the multipath.conf, the find_multipaths sections is as below, but we validate only invalid properties in the multipath.conf but not actual in memory multipath config value which we should do for Ubuntu.
https://github.com/hpe-storage/common-host-libs/blob/8a8f68fe84dbac8cd95a7902048a7ecf6b17fdb8/tunelinux/multipath.go#L270-L281

datamattsson:
Whatever we want to accomplish, is not being very effective, on-disk or in-memory. We need to ensure our change is bolted to the floor and persists.

raunakkumar:
Agree. This space needs few enhancements

datamattsson:
Where I'm getting at is that this does not solve the problem with Ubuntu which is the most pressing matter.

raunakkumar:
The PR does address the minimum viable configuration for 3par with `find_multipaths` set as no for Ubuntu for a new installation. This PR does not address existing multipath file and configuration. That needs more work for sure.

datamattsson:
> The PR does address the minimum viable configuration for 3par with `find_multipaths` set as no for Ubuntu for a new installation.

These files were not deployed on either @c-snell setup or the customer's.

raunakkumar:
> > The PR does address the minimum viable configuration for 3par with `find_multipaths` set as no for Ubuntu for a new installation.
> 
> These files were not deployed on either @c-snell setup or the customer's.

This change needs to be baked in the csi-driver image. So once this PR merges, I need to have a new csi driver build.

sneharai4:
@c-snell Is this being tested on centos and suse linux?

c-snell:
@sneharai4 The `multipath.conf` stanza that was submitted to @raunakkumar is for Ubuntu and iSCSI per [3PAR/Ubuntu documentation](https://support.hpe.com/hpesc/public/docDisplay?docId=c03961337&docLocale=en_US). RHEL 8.1, RHEL 7.4 and later incorporate the HPE required multipath settings in the built-in multipath module. There is no need to add 3PARdata device settings in /etc/multipath.conf, but the file must exist. If there are custom `multipath.conf` for SuSE, please work with @raunakkumar to include those during node conformance checks. 

I see the default SuSE/3PAR multipath is as follows:

```
device {
                vendor "3PARdata"
                product "VV"
                path_grouping_policy "group_by_prio"
                path_checker "tur"
                hardware_handler "1 alua"
                prio "alua"
                rr_weight "uniform"
                no_path_retry "18"
                failback "immediate"
        }
```

All OS configuration (iscsi, multipathing, etc) must be applied per OS during node conformance to avoid snowflake configurations. We need to ensure consistency across node conformance to include both 3PAR and Nimble configurations.

sneharai4:
@raunakkumar As discussed, since this needs more testing and code changes, let's hold on the changes for 2.0 release.

raunakkumar:
Closing as discussed offline. To be picked post 2.0

PR #157, opened 2021-06-21
CIDG-1395 : Changed the placement of logoutAndDeleteIscsiTarget() call

jyotsnalothe:
This is proposed fix for issue : CIDG-1395:  CSI 2.0.0 - Scale - Statefulset - iscsci | 50 Pod deletion took more that 25 minutes
Testing is still in progress. 

Please review it: 
@sneharai4 
@bhagyashree-sarawate 

Signed-off-by: jyotsna-l <jyotsna.l@hpe.com>

### DIFF
--- a/linux/device.go
+++ b/linux/device.go
@@ -1118,8 +1118,6 @@ func offlineScsiDevice(path string) (err error) {
 // DeleteDevice : delete the multipath device
 func DeleteDevice(dev *model.Device) (err error) {
 	log.Tracef("DeleteDevice called with %s", dev.SerialNumber)
-
-	isFC := isFibreChannelDevice(dev.Slaves)
 	// perform cleanup of the multipath device
 	if dev.SerialNumber == "" {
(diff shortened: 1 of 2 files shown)


linux/device.go
```diff
@@ -1139,6 +1141,12 @@ func DeleteDevice(dev *model.Device) (err error) {
 	if err != nil {
 		return err
 	}
+	if(!isFC){
+		err = logoutAndDeleteIscsiTarget(dev)
```
raunakkumar, line 1145: Why are we initiating iscsi logout and delete at multiple places in the same workflow.
Please remove this check as we already address it in `cleanupDeviceAndSlaves`

raunakkumar (changes requested):

linux/device.go
jyotsnalothe: As per the review comment, I have reverted the changes of logoutAndDeleteIscsiTarget function call from DeleteDevice function.

raunakkumar (approved):

PR #168, opened 2021-12-16
WIP: Remaplun fix by reading the wwid instead the vpdpath in case of 3PAR

bhagyashree-sarawate:
@raunakkumar , pavan-s-shanbhag@hpe.com
Still testing is in progress. so keeping the extra log lines  ..but raising the PR to get any early feedback for the fix and discussion

### DIFF
--- a/linux/device.go
+++ b/linux/device.go
@@ -621,7 +621,17 @@ func cleanupStaleScsiPaths(volume *model.Volume) (err error) {
 
 		// obtain current path serial using sysfs vpd80 page
 		var serialNumber string
-		serialNumber, err = getVpd80FromSysfs(h, c, t, l)
+		vendorName, err := getVendorFromSysfs(h, c, t, l)
+		if err != nil {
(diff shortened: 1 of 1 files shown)


linux/device.go
```diff
@@ -621,7 +621,17 @@ func cleanupStaleScsiPaths(volume *model.Volume) (err error) {
 
 		// obtain current path serial using sysfs vpd80 page
 		var serialNumber string
-		serialNumber, err = getVpd80FromSysfs(h, c, t, l)
+		vendorName, err := getVendorFromSysfs(h, c, t, l)
+		if err != nil {
```
raunakkumar, line 625: We should ignore the error and only care about the vendorName so that other vendors can proceed irrespective of the error.
please make the change in all occurrences

pavansshanbhag (approved):

raunakkumar (approved):
lgtm after the changes.

PR #169, opened 2022-03-11
CON-2295: Merge of multipath.conf is not comprehensive

adarshpratikhpe:


tunelinux/config/multipath.conf.generic
```diff
@@ -47,19 +47,6 @@ devices {
     device {
         vendor "3PARdata"
         product "VV"
-        path_grouping_policy "group_by_prio"
```
raunakkumar, line 50: Do we not have a default settings for 3par? Why are these removed

tunelinux/config/multipath.conf.upstream
```diff
@@ -49,19 +49,6 @@ devices {
     device {
         vendor "3PARdata"
         product "VV"
-        path_grouping_policy "group_by_prio"
```
raunakkumar, line 52: Same as above. Why are these removed?

tunelinux/config/multipath.conf.generic
adarshpratikhpe: For 3PAR we have different combination of suggested multipath configs depending on distro type.Depending on distro type some configs needs to be enabled or removed . Since there can be a lot of combination depending on distro type and version , only the common configs which will always be required for all distro types are kept in default settings. 

If the multipath.conf file is absent , default configuration will be applied first and then based on distro type the other recommended configs will be read from config.json and applied.

At present the Default section has only 2 configs i.e vendor and product type. In the next revisions when we add the section for other distro Type, we will figure out the common configs and add it in the default settings in this file.

tunelinux/config/multipath.conf.upstream
adarshpratikhpe: Same as multipath.conf.generic file.

tunelinux/config/multipath.conf.generic
raunakkumar: Thanks for the clarification. We need to make sure all the different distros have been tested.

pavansshanbhag (changes requested):

tunelinux/multipath.go
```diff
@@ -328,6 +329,23 @@ func SetMultipathRecommendations() (err error) {
 			return err
 		}
 	}
+	// Get current recommendations
+	recommendations, err := GetMultipathRecommendations()
+	if err != nil {
+		return err
+	}
+	if len(recommendations) == 0 {
+		log.Info("no recommendations found for multipath.conf settings")
```
pavansshanbhag, line 338: Missed this one - can you make this a warning level log

adarshpratikhpe: Done. Log level changed to Warning.

raunakkumar (approved):

PR #177, opened 2023-02-08
Update cmd.go

NoamAsso:
fixing an issue where the timeout for creating the filesystem is not 300 sec causing it to fail when creating large size volumes. this is because ExecCommandOutputWithTimeout receives a parameter "timeout" but then always runs execCommandOutputWithTimeout with "defaultTimeout" instead of passing the "timeout"

### DIFF
--- a/util/cmd.go
+++ b/util/cmd.go
@@ -28,7 +28,7 @@ import (
 )
 
 const (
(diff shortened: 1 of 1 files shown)


sijeesh:
@NoamAsso , This change is already submited in #175.

PR #180, opened 2024-02-12
CON-2137:File System Corruption Fix

AnushaY1916:


dileepds (commented):
1. We can have a common method to find the Filesystem type which can be used for Filesystem corruption checking and Filesystem repair.
2. I think the method executeFilesystemRepair can be used in fsck repair with additional code for handling the exit status.

AnushaY1916:
> 1. We can have a common method to find the Filesystem type which can be used for Filesystem corruption checking and Filesystem repair.

These methods are kept separated because we detect the file system corruption first and then try to repair it only if the user has opted to repair it.

AnushaY1916:
> 2\. I think the method executeFilesystemRepair can be used in fsck repair with additional code for handling the exit status.

fsck has many return codes that need to be handled each one differently and it will be confusing to handle all of the file systems together.

dileepds (approved):
LGTM

datamattsson (approved):
LGTM

PR #186, opened 2024-06-24
CON2508: Handling the Cleanup  of stale multipath devices

AnushaY1916:
The following functionalities are implemented in phase 3:

1. Unmount all references to the device, bind mounts, and real mounts.
2. Kill processes using the mount points and devices
3. Remove block devices
4. Flush the multipath map

made improvements for the previous  phase's implementation as well


datamattsson (changes requested):
Can we fix the spelling of `func KillProcessesUisngMountPoints`?

datamattsson (approved):

PR #187, opened 2024-07-17
Legacy fixes

datamattsson:
- This fixes prevention of the CSI driver from starting as multipath commands writes non-parseable content on stderr for modern Linux distributions.


dileepds (approved):
LGTM

PR #192, opened 2024-12-19
Upgraded crypto

sijeesh:
Upgraded crypto version to fix a critical vulnerability


go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/hpe-storage/common-host-libs
 
-go 1.19
+go 1.20
```
datamattsson, line 3: I don't think all driver components and sidecars compile with 1.20, have you verified?

sijeesh: Successfully compiled all other repos with this common host lib and go versions. Tested all the latest images with a local helm chart

dileepds (approved):
LGTM

datamattsson (approved):

PR #193, opened 2025-03-04
CON-3240 Implement the csi-driver changes in CreateVolume to invoke t…

diganta-001:
…he CSP.

### DIFF
--- a/model/types.go
+++ b/model/types.go
@@ -457,3 +457,9 @@ type ProcMount struct {
 	FileSystem string `json:"FileSystem,omitempty"`
 	Options    string `json:"Options,omitempty"`
 }
+
+type MountOption struct {
+        UserSquashing string
+        NfsVersion    string
+        ReadOnly      bool
+}


model/types.go
```diff
@@ -457,3 +457,9 @@ type ProcMount struct {
 	FileSystem string `json:"FileSystem,omitempty"`
 	Options    string `json:"Options,omitempty"`
 }
+
+type MountOption struct {
+        UserSquashing string
+        NfsVersion    string
+        ReadOnly      bool
+}
```
datamattsson, line 465: `UserSquashing` should be handled by the CSP parameters.
`NfsVersion` should server side should be handled by the CSP parameters, client side in the StorageClass mountOptions.
`ReadOnly` should be handled by the CSP parameters.

Please close this PR.

datamattsson (changes requested):
Delete this PR, this is the wrong place.

PR #198, opened 2025-05-12
Fix for iSCSI VSTs

datamattsson:
Fixes for detaching volume scoped targets (VST) for iSCSI.
- Enumerate each target
- Pass if targets nodes and sessions are missing
- Exits if mpath device is busy flushing caches from the umount

Previous behavior enumerated all connected targets and logged out all targets on first pass. Additional passes fail and delays detaching significantly (10 retries) and confuses the external attacher which lead to 5+ minutes long process for detaching a device and lingering iSCSI sessions.

Passes CSI e2e tests for HPE iSCSI GSTs (stock Nimble and 3PAR) and VSTs (old Nimbles and 3rd party).

### DIFF
--- a/linux/iscsi.go
+++ b/linux/iscsi.go
@@ -54,6 +54,8 @@ const (
 	PingTimeout             = 5 * time.Second
 	DefaultIscsiPort        = 3260
 	iscsiSessionDir         = "/sys/class/iscsi_session"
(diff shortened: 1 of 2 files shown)


dileepds (approved):
LGTM

AnushaY1916 (approved):
LGTM

PR #226, opened 2026-07-09
CON-4800 Increase Unix socket timeout for multipathd communication fr…

diganta-001:
…om node/init container, one additional task will be to  add env variable for helm

### DIFF
--- a/tunelinux/config/multipath.conf.generic
+++ b/tunelinux/config/multipath.conf.generic
@@ -1,6 +1,7 @@
 defaults {
     user_friendly_names yes
     find_multipaths     no
+    uxsock_timeout      300000
 }
 blacklist {
     devnode "^(ram|raw|loop|fd|md|dm-|sr|scd|st)[0-9]*"
--- a/tunelinux/config/multipath.conf.upstream
+++ b/tunelinux/config/multipath.conf.upstream
@@ -1,7 +1,7 @@
 defaults {
     user_friendly_names yes
     find_multipaths     no
-    uxsock_timeout      10000
+    uxsock_timeout      300000
 }
 blacklist {
     devnode "^(ram|raw|loop|fd|md|dm-|sr|scd|st)[0-9]*"
--- a/tunelinux/multipath.go
+++ b/tunelinux/multipath.go
@@ -24,6 +24,18 @@ import (
 const (
 	multipath = "multipath"
 
+	// uxsockTimeoutParam is the multipath.conf defaults-section key that controls
+	// the unix domain socket timeout (in milliseconds) used by multipathd.
+	uxsockTimeoutParam = "uxsock_timeout"
+
+	// uxsockTimeoutRecommended is the default recommended value for uxsock_timeout. Keep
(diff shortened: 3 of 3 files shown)


tunelinux/config/multipath.conf.generic
```diff
@@ -1,4 +1,5 @@
 defaults {
+    uxsock_timeout      300000
```
datamattsson: Can we make this change in multipath.conf.upstream too and make sure the keywords are in the same place etc.

datamattsson (changes requested):

kalikivayi (changes requested):
I have reviewed manually as well as with the AI MCP. I do not have any manual comments, but AI has given few comments. Please address them.

[con-4800_review.md](https://github.com/user-attachments/files/29883756/con-4800_review.md)

tunelinux/config/multipath.conf.generic
diganta-001: file added

diganta-001: > Can we make this change in multipath.conf.upstream too and make sure the keywords are in the same place etc

file added , thanks

diganta-001:
> I have reviewed manually as well as with the AI MCP. I do not have any manual comments, but AI has given few comments. Please address them.
> 
> [con-4800_review.md](https://github.com/user-attachments/files/29883756/con-4800_review.md)

Here is the Copilot and developer response 
1 (defaults section) / 4 (no reload) — Not valid concerns. setMultipathRecommendations always creates the defaults{} section if missing and always reaches the uxsock_timeout reconciliation block. SetMultipathRecommendations unconditionally calls multipathd reconfigure after saving, so the value takes effect immediately on a running daemon.
2 (justify 300s) — Will add a doc comment; the 5-minute value will also be captured in the docs, so no action needed in code itself.
3  (upper bound / clamp) — Not adding a hard max. Capping it in code is actually risky — a hardcoded ceiling could itself become the wrong value for some environment/scale, and would require another code change to adjust. The env var (MULTIPATH_UXSOCK_TIMEOUT) already gives admins full control to tune it themselves if 300000ms isn't right for their setup, so this is intentionally left open rather than artificially bounded.
5 (const/template duplication) + missing tests — tunelinux/ has no existing test coverage for any function today (verified — zero _test.go files, no git history), so adding new tests here isn't restoring parity with an existing convention. Treating as a follow-up, not a blocker for this PR.

kalikivayi (approved):
LGTM

rohit-balakrishna (approved):
@diganta-001 , please share the e2e test results with this patch and the GetSnapshots change. Thanks

rohit-balakrishna:
> @diganta-001 , please share the e2e test results with this patch and the GetSnapshots change. Thanks

CSI E2E after applying both changes in PR226 and PR229 to the CSI Driver and 3par CSP passes and results are:

Ran 99 of 7907 Specs in 6289.825 seconds
PASS
Created /data/csi-e2e-vpl6f-26-07-14-06-43-03.xml.gz
`
@datamattsson please approve.

datamattsson (approved):

PR #227, opened 2026-07-09
CON-4772 optimize GetMultiPathDevices function

prashantkumars289:
To optimize memory usage in the GetMultipathDevices function, the following changes were implemented:

1) Directly unmarshaled the output string without converting it to a byte slice.

2) Instead of logging the entire multipathDevices, logged only the counts: the length of multipathDevices, unhealthyDeviceCount, and orphanDeviceCount.

Design Document: https://confluence.storage.hpecorp.net/spaces/ESI/pages/682047152/CON-4772+-+Fix+hpe-csi-node+OOM+kill+issue 

This also contains a fix for the Jira https://jira.storage.hpecorp.net/browse/CON-3096

### DIFF
--- a/tunelinux/multipath.go
+++ b/tunelinux/multipath.go
@@ -393,50 +393,52 @@ func GetMultipathDevices() (multipathDevices []model.MultipathDevice, err error)
 		return nil, fmt.Errorf("Failed to get the multipath devices due to the error: %s", err.Error())
 	}
 
-	if out != "" {
-		multipathJson := new(model.MultipathInfo)
-		//err = json.Unmarshal([]byte(out), multipathJson)
-		decoder := json.NewDecoder(strings.NewReader(out))
-		err = decoder.Decode(multipathJson)
-		if err != nil {
-			return nil, fmt.Errorf("Invalid JSON output of multipathd command: %s", err.Error())
+	out = strings.TrimSpace(out)
+	if out == "" {
+		log.Infof("No multipath devices found")
+		return nil, nil
+	}
+
+	multipathJson := new(model.MultipathInfo)
+	decoder := json.NewDecoder(strings.NewReader(out))
+	err = decoder.Decode(multipathJson)
+	if err != nil {
+		return nil, fmt.Errorf("Invalid JSON output of multipathd command: %s", err.Error())
+	}
+	var extra interface{}
+	if err = decoder.Decode(&extra); err != io.EOF {
+		if err == nil {
+			err = fmt.Errorf("unexpected data after JSON object")
 		}
-		var extra interface{}
-		if err = decoder.Decode(&extra); err != io.EOF {
-			if err == nil {
-				err = fmt.Errorf("unexpected data after JSON object")
-			}
-			return nil, fmt.Errorf("Invalid JSON output of multipathd command: %s", err.Error())
-		}
+		return nil, fmt.Errorf("Invalid JSON output of multipathd command: %s", err.Error())
+	}
 
-		multipathDevices = make([]model.MultipathDevice, 0, len(multipathJson.Maps))
-		unhealthyDeviceCount := 0
-		orphanDeviceCount := 0
+	multipathDevices = make([]model.MultipathDevice, 0, len(multipathJson.Maps))
+	unhealthyDeviceCount := 0
+	orphanDeviceCount := 0
 
-		for _, mapItem := range multipathJson.Maps {
-			if len(mapItem.Vend) > 0 && isSupportedDeviceVendor(linux.DeviceVendorPatterns, mapItem.Vend) {
-				if mapItem.Paths < 1 && mapItem.PathFaults > 0 {
-					mapItem.IsUnhealthy = true
-					unhealthyDeviceCount++
-					log.Tracef("Known unhealthy multipath device: %s", mapItem.Name)
-				}
-				multipathDevices = append(multipathDevices, mapItem)
-				continue
-			}
-
-			// residual non-functional multipath devices
-			if len(mapItem.PathGroups) == 0 && mapItem.Queueing == "off" && mapItem.Features == "0" {
+	for _, mapItem := range multipathJson.Maps {
+		if len(mapItem.Vend) > 0 && isSupportedDeviceVendor(linux.DeviceVendorPatterns, mapItem.Vend) {
+			if mapItem.Paths < 1 && mapItem.PathFaults > 0 {
 				mapItem.IsUnhealthy = true
 				unhealthyDeviceCount++
-				orphanDeviceCount++
-				log.Tracef("Unknown orphan multipath device without path_groups: %s", mapItem.Name)
-				multipathDevices = append(multipathDevices, mapItem)
+				log.Tracef("Known unhealthy multipath device: %s", mapItem.Name)
 			}
+			multipathDevices = append(multipathDevices, mapItem)
+			continue
 		}
-		log.Infof("Found %d multipath devices (%d unhealthy, %d orphan)", len(multipathDevices), unhealthyDeviceCount, orphanDeviceCount)
-		return multipathDevices, nil
+
+		// residual non-functional multipath devices
+		if len(mapItem.PathGroups) == 0 && mapItem.Queueing == "off" && mapItem.Features == "0" {
+			mapItem.IsUnhealthy = true
+			unhealthyDeviceCount++
+			orphanDeviceCount++
+			log.Tracef("Unknown orphan multipath device without path_groups: %s", mapItem.Name)
+			multipathDevices = append(multipathDevices, mapItem)
+		}
 	}
-	return nil, fmt.Errorf("Invalid multipathd command output received")
+	log.Infof("Found %d multipath devices (%d unhealthy, %d orphan)", len(multipathDevices), unhealthyDeviceCount, orphanDeviceCount)
+	return multipathDevices, nil
 }
 
 func isSupportedDeviceVendor(deviceVendors []string, vendor string) bool {
--- a/tunelinux/multipath.go
+++ b/tunelinux/multipath.go
@@ -6,6 +6,7 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"io"
 	"io/ioutil"
 	"os"
 	"os/exec"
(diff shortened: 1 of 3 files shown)


prashantkumars289:
# PR 227 Review: CON-4772 optimize GetMultiPathDevices function

PR: https://github.com/hpe-storage/common-host-libs/pull/227

Review date: 2026-07-19

## Summary

PR #227 updates multipath device discovery to reduce memory pressure and avoid logging full device payloads from `multipathd show multipaths json`. The latest local/PR state also includes follow-up handling for transient `multipathd` timeout output and removes deprecated `io/ioutil` usage from `tunelinux/multipath.go`.

Changed files reported by the MCP PR context:

- `linux/multipath.go`
- `tunelinux/multipath.go`
- `tunelinux/multipath_test.go`

## CSI Storage MCP Review

Tool used: `CSI_Storage_MCPServer` / `mcp_csi_storage_m_review_pr`

MCP result:

- Critical: 1
- Warning: 0
- Info: 0
- Suggestion: 0

MCP finding reviewed:

- `ERR-001` in `tunelinux/multipath_test.go`: MCP reported an unchecked error after `devices, err := parseMultipathDevices(...)` in `TestParseMultipathDevicesClassifiesUnhealthyAndOrphan`.
- Local review result: this appears to be a false positive. The test checks `if err != nil { t.Fatalf(...) }` immediately after the assignment before inspecting `devices`, so there is no nil/panic risk from proceeding after a failed parse.

## Review Findings

No blocking issues found in the current local implementation.

## Implemented Behavior Contract

The requested parse/classification contract is implemented in `parseMultipathDevices(out string)` and covered by table-driven tests in `tunelinux/multipath_test.go`.

| Input | Result |
|---|---|
| transient `timeout` / `receiving packet` (even rc=0) | `(nil, ErrMultipathTimeout)` |
| empty / whitespace-only | `(nil, nil)` — no devices |
| invalid / trailing-garbage JSON | `(nil, error)` |
| valid JSON | `(devices, nil)` |

Evidence:

- `linux.IsMultipathTimeoutError` is exported and reused by `tunelinux.parseMultipathDevices`, keeping the timeout string classification consistent with existing multipath helpers.
- `GetMultipathDevices` retries `ErrMultipathTimeout` up to `multipathTimeoutMaxTries` before returning a wrapped error, so callers can still detect it with `errors.Is(err, ErrMultipathTimeout)`.
- Empty and whitespace-only command output returns `(nil, nil)`, preserving the intended "no devices" behavior and avoiding a node-init failure for empty output.
- Invalid JSON and trailing data are still rejected, including the explicit `unexpected data after JSON object` path.
- Valid JSON is decoded with `json.NewDecoder(strings.NewReader(out))`, avoiding the previous full `[]byte` copy allocation.

## Additional Review Notes

- Deprecated `io/ioutil` usage has been removed from `tunelinux/multipath.go`; `os.ReadFile` is used for reading `linux.MultipathConf`.
- Logging now reports counts with `Found %d multipath devices (%d unhealthy, of which %d are orphan)`, which clarifies that orphan devices are included in the unhealthy count.
- The parser preallocates `multipathDevices` using `len(multipathJson.Maps)`, reducing append growth while filtering unsupported devices.
- The retry path intentionally treats the timeout/receiving-packet symptom as retriable rather than mapping it to `(nil, nil)`, which avoids silently skipping cleanup when multipathd is temporarily busy.

## Residual Integration Risk

After retries are exhausted, `GetMultipathDevices` returns a wrapped `ErrMultipathTimeout`. Any caller that treats all errors as fatal can still fail during a persistent multipathd busy/reload window. CSI driver node-init should handle `errors.Is(err, tunelinux.ErrMultipathTimeout)` as a non-fatal skip-and-retry condition when this common-host-libs update is vendored.

## Validation

- `go test ./tunelinux` passed locally.

## Recommendation

Approve PR #227 from the common-host-libs side. The MCP-reported critical item is not actionable in the current code, and the requested timeout/empty/invalid/valid parsing contract is implemented and tested.

rohit-balakrishna:
Fix the commit message for commit: 4c066868278f30ea2622ae3faa5e16d723bdd822

tunelinux/multipath.go
```diff
@@ -367,34 +368,52 @@ func GetMultipathDevices() (multipathDevices []model.MultipathDevice, err error)
 		return nil, fmt.Errorf("Failed to get the multipath devices due to the error: %s", err.Error())
 	}
 
-	if out != "" {
-		multipathJson := new(model.MultipathInfo)
-		err = json.Unmarshal([]byte(out), multipathJson)
-		if err != nil {
-			return nil, fmt.Errorf("Invalid JSON output of multipathd command: %s", err.Error())
+	out = strings.TrimSpace(out)
```
rohit-balakrishna, line 422: In this file linux/multipath.go, after calling ExecCommandOutput(multipathd ...) function, we also invoke isMultipathTimeoutError() with the response as shown below: 

`	out, _, err := util.ExecCommandOutput(multipathd, showPathsFormat)
	if err != nil {
		log.Warnf("multipathdShowCmd: error %v with args %v", err, showPathsFormat)
		return nil, err
	}
	// rc can be 0 on the below error conditions as well
	if isMultipathTimeoutError(out) {
		err = fmt.Errorf("failed to get multipathd %v, out %s", showPathsFormat, out)
		log.Warn(err.Error())
		return nil, err
	}
`
Whereas we dont do that here. We can discuss this further. Will it be caught under `unexpected data after JSON object` ?

tunelinux/multipath.go
```diff
@@ -6,6 +6,7 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"io"
 	"io/ioutil"
```
kalikivayi: Can we use `io` at all places instead of `io/ioutil`? `ioutil` is depricated.

tunelinux/multipath.go
```diff
@@ -367,34 +368,52 @@ func GetMultipathDevices() (multipathDevices []model.MultipathDevice, err error)
...
 
-	if out != "" {
-		multipathJson := new(model.MultipathInfo)
-		err = json.Unmarshal([]byte(out), multipathJson)
-		if err != nil {
-			return nil, fmt.Errorf("Invalid JSON output of multipathd command: %s", err.Error())
+	out = strings.TrimSpace(out)
+	if out == "" {
+		log.Infof("No multipath devices found")
+		return nil, nil
```
kalikivayi, line 425: Should we return `nil, error` here, instead of `nil, nil`?

tunelinux/multipath.go
```diff
@@ -367,34 +368,52 @@ func GetMultipathDevices() (multipathDevices []model.MultipathDevice, err error)
...
 			}
+			multipathDevices = append(multipathDevices, mapItem)
+			continue
+		}
+
+		// residual non-functional multipath devices
+		if len(mapItem.PathGroups) == 0 && mapItem.Queueing == "off" && mapItem.Features == "0" {
+			mapItem.IsUnhealthy = true
+			unhealthyDeviceCount++
+			orphanDeviceCount++
```
kalikivayi, line 464: An orphan device is also treated as unhealthy device. In the line 415, users might get confused that that there are `x` unhealthy and `y` orphan devices. But in reality, there are total `x` unhealthy devices and out of `x` unhealthy devices `y` are orphaned.

kalikivayi (changes requested):
Please address review comments

tunelinux/multipath.go
prashantkumars289: Implemented retry on multipath command timeout

tunelinux/multipath.go
prashantkumars289: Done

tunelinux/multipath.go
prashantkumars289: Returning nil, the error will fail node-init, which will lead to the CSI-node crashing. To avoid that, we have implemented the following

  | Input | Result |
  |---|---|
  | transient `timeout` / `receiving packet` (even rc=0) | `(nil, ErrMultipathTimeout)` |
  | empty / whitespace-only | `(nil, nil)` — no devices |
  | invalid / trailing-garbage JSON | `(nil, error)` |
  | valid JSON | `(devices, nil)` |

tunelinux/multipath.go
prashantkumars289: Improved the log statement

log.Infof("Found %d multipath devices (%d unhealthy, of which %d are orphan)", len(multipathDevices), unhealthyDeviceCount, orphanDeviceCount)

tunelinux/multipath.go
```diff
@@ -361,40 +365,77 @@ func GetMultipathDevices() (multipathDevices []model.MultipathDevice, err error)
...
+				return nil, fmt.Errorf("failed to get the multipath devices after %d attempts: %w", multipathTimeoutMaxTries, err)
+			}
+			log.Warnf("Transient multipathd timeout while getting multipath devices; retrying attempt %d of %d", try+1, multipathTimeoutMaxTries)
+			time.Sleep(multipathTimeoutRetrySleep)
+			continue
+		}
+		if cmdErr != nil {
+			return nil, fmt.Errorf("failed to get the multipath devices due to the error: %s", cmdErr.Error())
+		}
+		return multipathDevices, err
```
kalikivayi: Why are we returning err here. Should we return nil instead of err

kalikivayi (commented):
Please address review comment

tunelinux/multipath.go
prashantkumars289: If there is any error in the multipathd command, we return that error. 

The csi-driver doesn't mark timeout errors as fatal, whereas other errors are. For errors other than timeout errors, nodeinit will fail, and the csi-node pod will crash.

kalikivayi: In that case why are we not checking for err != nil, as done for cmdErr.
Can we make line 385 as below:
if cmdErr != nil || err != nil{
    //return nil, err
}
return multipathDevices, nil

kalikivayi (commented):
Please address review comment

tunelinux/multipath.go
prashantkumars289: Added error handling for better readability.

kalikivayi (approved):
LGTM

rohit-balakrishna (approved):
Ran E2E: CSI_E2E_CONFIG=primera-iscsi-block-e2e and results are below:

Ran 99 of 7907 Specs in 7014.466 seconds
PASS
Created /data/csi-e2e-gthkt-26-07-20-10-53-33.xml.gz
`

rohit-balakrishna:
@datamattsson , @dileepds , please review. Thanks

datamattsson (approved):

PR #229, opened 2026-07-10
CON-4709 Fix the GetVolumeSnapshot issue causing OOM panic restarts o…

diganta-001:
…f 3par CSP , we have added a new query params in  getsanpshot API , it enables consumer CSP to use same api for different based on mode provided

### DIFF
--- a/storageprovider/csp/container_storage_provider.go
+++ b/storageprovider/csp/container_storage_provider.go
@@ -767,8 +767,13 @@ func (provider *ContainerStorageProvider) GetVolumes() ([]*model.Volume, error) 
 	return volumes, err
 }
 
-// GetSnapshots returns all of the snapshots for the given source volume from the CSP
-func (provider *ContainerStorageProvider) GetSnapshots(volumeID string) ([]*model.Snapshot, error) {
+// GetSnapshots returns all of the snapshots for the given source volume from the CSP.
+// mode is variadic (instead of a plain optional parameter) so this method's signature
+// stays backward compatible for callers that don't care about mode - e.g. CSI's
+// ListSnapshots and DeleteVolume/deleteVolume paths call GetSnapshots(volumeID) with no
+// mode at all. Only mode[0] is read; it's passed through as the "mode" query param to the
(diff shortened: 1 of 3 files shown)


storageprovider/csp/container_storage_provider.go
```diff
@@ -768,7 +768,7 @@ func (provider *ContainerStorageProvider) GetVolumes() ([]*model.Volume, error)
 }
 
 // GetSnapshots returns all of the snapshots for the given source volume from the CSP
-func (provider *ContainerStorageProvider) GetSnapshots(volumeID string) ([]*model.Snapshot, error) {
+func (provider *ContainerStorageProvider) GetSnapshots(volumeID string, mode ...string) ([]*model.Snapshot, error) {
```
kalikivayi, line 776: When we are taking the first element for mode, what is the use of defining the mode as `Variadic`? Can you add the reason as a comment?

kalikivayi (changes requested):
Please address review comment

storageprovider/csp/container_storage_provider.go
diganta-001: from csi delete and list snapshot call this api, for delete flow csp has a different workflow , to distinguish this we have kept it like this so that list snapshot can call witout the last arg

kalikivayi (approved):
LGTM

storageprovider/csp/container_storage_provider.go
rohit-balakrishna, line 776: please update the comment about the variadic mode argument.

rohit-balakrishna:
CSI E2E after applying both changes in PR226 and PR229 to the CSI Driver and 3par CSP passes and results are:

Ran 99 of 7907 Specs in 6289.825 seconds
PASS
Created /data/csi-e2e-vpl6f-26-07-14-06-43-03.xml.gz
`
The CSP PR for the corresponding changes are already merged. @datamattsson please approve.

rohit-balakrishna (approved):

datamattsson (approved):

PR #231, opened 2026-07-13
Log warning on nguid read failure

msitte:
Log a warning when failing to read nguid from sysfs path instead of returning an error. A missing NGUID on one NVMe namespace should not terminate device discovery.


msitte:
Issue observed on server with NVMe boot device enumerated as /dev/nvme0n1.

datamattsson:
Thank you for your finding. @rohit-balakrishna can we have someone review this?

@msitte the PRs that are submitted to the HPE CSI Driver repos need to be signed. Please rebase and sign your commits.

datamattsson:
Fixed in #233

PR #235, opened 2026-07-31
CON-4193 VM stop and subsequent start does not work in case of primary storage array failure chl changes

rashi-jain1:
JIRA Link: https://jira.storage.hpecorp.net/browse/CON-4193

### DIFF
--- a/linux/iscsi.go
+++ b/linux/iscsi.go
@@ -222,60 +222,26 @@ func loginToVolume(volume *model.Volume) (err error) {
 }
 
 // HandleIscsiDiscovery performs iscsi target discovery and create sessions as required.
+// NOTE: This function handles a single backend (primary or secondary). The caller
+// (rescanLoginVolume in device.go) is responsible for iterating over secondaries.
 func HandleIscsiDiscovery(volume *model.Volume) (err error) {
 	log.Tracef(">>>>> HandleIscsiDiscovery for volume %s, lun %s", volume.SerialNumber, volume.LunID)
 	defer log.Tracef("<<<<< HandleIscsiDiscovery")
 
-	// Do iscsi discovery for Primary Backend
-
-	var primaryVolObj *model.Volume
-	primaryVolObj = &model.Volume{}
-	primaryVolObj.LunID = volume.LunID
-	primaryVolObj.Iqns = volume.Iqns
-	primaryVolObj.TargetScope = volume.TargetScope
-	primaryVolObj.DiscoveryIPs = volume.DiscoveryIPs
-	primaryVolObj.Chap = volume.Chap
-	primaryVolObj.ConnectionMode = volume.ConnectionMode
-	primaryVolObj.SerialNumber = volume.SerialNumber
-
 	// During APP failover the primary array may be offline and its target IQNs
-	// will be absent from the publish context. Only skip primary discovery when there are
-	// genuinely no targets — any other failure means the primary is faulty
+	// will be absent from the publish context. Only skip discovery when there are
+	// genuinely no targets — any other failure means the array is faulty
 	// and we should surface the error.
 	// Note: targetNames:"" parses as []string{""} so we must filter empty strings.
-	if len(filterEmptyTargets(primaryVolObj.TargetNames())) == 0 {
-		log.Warnf("primary backend has no target IQNs, skipping primary iSCSI discovery (APP failover scenario)")
-	} else {
-		err = handleIscsiDiscoveryForBackend(primaryVolObj, true)
-		if err != nil {
-			return err
-		}
+	if len(filterEmptyTargets(volume.TargetNames())) == 0 {
+		log.Warnf("no target IQNs for volume %s, skipping iSCSI discovery (APP failover scenario)", volume.SerialNumber)
+		return nil
 	}
 
-	secondaryBackends := util.GetSecondaryBackends(volume.SecondaryArrayDetails)
-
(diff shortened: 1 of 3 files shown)


linux/iscsi.go
```diff
@@ -238,11 +238,20 @@ func HandleIscsiDiscovery(volume *model.Volume) (err error) {
...
+	} else {
+		err = handleIscsiDiscoveryForBackend(primaryVolObj, true)
+		if err != nil {
+			return err
+		}
 	}
+
 	secondaryBackends := util.GetSecondaryBackends(volume.SecondaryArrayDetails)
 
 	for _, secondaryLunInfo := range secondaryBackends {
```
diganta-001: Question (existing structure, non-blocking): secondaries are already handled by rescanLoginVolume calling HandleIscsiDiscovery per backend (device.go L439), yet HandleIscsiDiscovery has its own secondary loop too. Since SecondaryArrayDetails isn't copied into the per-backend objects, that inner loop is a no-op today - so
the new `if err != nil { return err }` inside it won't run here. @rohit-balakrishna

rohit-balakrishna: Agree, we need to understand this better. This seems to be dead code. @rashi-jain1 , please check and fix.

linux/iscsi.go
```diff
@@ -238,11 +238,20 @@ func HandleIscsiDiscovery(volume *model.Volume) (err error) {
...
-
-	if err != nil {
-		return err
+	// During APP failover the primary array may be offline and its target IQNs
+	// will be absent from the publish context. Only skip primary discovery when there are
+	// genuinely no targets — any other failure means the primary is faulty
+	// and we should surface the error.
+	// Note: targetNames:"" parses as []string{""} so we must filter empty strings.
+	if len(filterEmptyTargets(primaryVolObj.TargetNames())) == 0 {
+		log.Warnf("primary backend has no target IQNs, skipping primary iSCSI discovery (APP failover scenario)")
```
rohit-balakrishna: Dont we need similar change for FC ?

linux/iscsi.go
```diff
@@ -258,6 +267,10 @@ func HandleIscsiDiscovery(volume *model.Volume) (err error) {
 		secondaryVolObj.SerialNumber = volume.SerialNumber
 
 		err = handleIscsiDiscoveryForBackend(secondaryVolObj, true)
+		if err != nil {
+			log.Errorf("secondary backend discovery failed for targets %v: %s", secondaryVolObj.TargetNames(), err.Error())
+			return err
+		}
 	}
 	return nil
```
rohit-balakrishna: If primary is skipped and seccondary is empty, then we return nil, which is not correct. if we fix the dead code issue, this should also get handled.

linux/iscsi.go
```diff
@@ -258,6 +267,10 @@ func HandleIscsiDiscovery(volume *model.Volume) (err error) {
 		secondaryVolObj.SerialNumber = volume.SerialNumber
 
 		err = handleIscsiDiscoveryForBackend(secondaryVolObj, true)
```
rohit-balakrishna: For both Primary and secondary, we are passing `isPrimaryBackend` as true. Do we need this argument?

rohit-balakrishna:
The function `rescanIscsiHosts` was changed to skip non existent hostScan paths, but the subsequent WriteFile failures are not skipped. I have updated the review md file with the analysis. 
[CON-4193-both-prs-approval-summary.md](https://github.com/user-attachments/files/30646237/CON-4193-both-prs-approval-summary.md)
Please find attached above. Please check

linux/iscsi.go
rashi-jain1: No, FC doesn't need the similar change because the FC discovery path (rescanLoginVolumeForBackend) is structurally different from iSCSI. There's no per-target session login for FC, so a down primary array doesn't produce the login/rescan hard-error that the iSCSI primary path hits.

linux/iscsi.go
rashi-jain1: Yes, Removed the dead inner loop. HandleIscsiDiscovery now handles exactly the single backend it's given, and secondary iteration stays in rescanLoginVolume.

linux/iscsi.go
rashi-jain1: Fixed

linux/iscsi.go
rashi-jain1: No, it was only used in a trace log not in the logic, and after removing the dead secondary loop there's a single caller so removed the parameter.

rashi-jain1:
> The function `rescanIscsiHosts` was changed to skip non existent hostScan paths, but the subsequent WriteFile failures are not skipped. I have updated the review md file with the analysis. [CON-4193-both-prs-approval-summary.md](https://github.com/user-attachments/files/30646237/CON-4193-both-prs-approval-summary.md) Please find attached above. Please check

Have Addressed the Required Once. Thanks

diganta-001 (approved):
LGTM

rohit-balakrishna:
Results of E2E: CSI_E2E_CONFIG=primera-iscsi-block-e2e

Ran 99 of 7907 Specs in 4907.822 seconds
PASS
Created /data/csi-e2e-lm9nw-26-08-03-12-18-26.xml.gz

@datamattsson , @dileepds , please review.

rohit-balakrishna (approved):

dileepds (approved):
LGTM

PR #237, opened 2026-08-03
CON-4777: fix csi node init crash on restart

prashantkumars289:
The following changes are made through this PR

1) Make defaultTimeout env-overridable through the HPE_CMD_TIMEOUT_SECONDS environment variable.

2) Set the multipath command executor timeout to 5 seconds more than multipath ux_sock_timeout. 

3) Skip multipath reconfigure if there is no change in multipath.conf

Reference: https://jira.storage.hpecorp.net/browse/CON-4777

### DIFF
--- a/linux/device.go
+++ b/linux/device.go
@@ -876,6 +876,7 @@ func getFcTargetWwpns(volume *model.Volume) []string {
 
 // find if an existing lun has been remapped with new lun provided
 // if found, remove old paths and rescan new lun paths
+//
 //nolint:gocyclo
 func handleRemappedLun(volume *model.Volume) (err error) {
 	log.Tracef(">>> handleRemappedLun called with volume %s serial %s lun %s", volume.Name, volume.SerialNumber, volume.LunID)
@@ -1559,7 +1560,7 @@ func RescanForCapacityUpdates(devicePath string) error {
 		devicePath = strings.TrimPrefix(devicePath, "/dev/mapper/")
 		// reload multipath map to apply new size
 		args = []string{"resize", "map", devicePath}
-		out, _, err := util.ExecCommandOutput("multipathd", args)
+		out, _, err := multipathdExecCommandOutput("multipathd", args, MultipathdCommandTimeout())
 		if err != nil {
 			return err
 		}
--- a/linux/device.go
+++ b/linux/device.go
@@ -1422,7 +1422,7 @@ func DeleteDevice(dev *model.Device) (err error) {
 // flush the device buffers
 func flushbufs(dev *model.Device) error {
 	log.Tracef("flushbufs called for %+v", dev)
-	if dev == nil && dev.AltFullPathName == "" {
+	if dev == nil || dev.AltFullPathName == "" {
 		return fmt.Errorf("device.AltFullPathName %+v not present to perform flushbufs", dev)
 	}
 	args := []string{"--flushbufs", dev.AltFullPathName}
--- a/linux/multipath.go
+++ b/linux/multipath.go
@@ -5,8 +5,10 @@ package linux
 import (
 	"fmt"
 	"io/ioutil"
+	"os"
 	"path/filepath"
 	"regexp"
+	"strconv"
 	"strings"
 	"sync"
 	"time"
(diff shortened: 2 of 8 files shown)


prashantkumars289:
# PR Review — hpe-storage/common-host-libs #237

**Title:** Con 4777 fix csi node init crash on restart
**Author:** Prashant Kumar
**Branch:** `CON-4777_fix-csi-node-init-crash-on-restart` → `master`
**Jira:** [CON-4777](https://jira.storage.hpecorp.net/browse/CON-4777)
**Scope:** 8 files changed, +358 / −107
**Reviewed with:** CSI_Storage_MCPServer (`review_pr`, `get_pr_context`)

---

## Summary of Changes

The PR addresses a CSI node driver crash observed on pod/node restart. It makes three functional changes plus supporting refactors:

1. **Configurable command timeout** — `defaultTimeout` in `util/cmd.go` is now overridable via the `HPE_CMD_TIMEOUT_SECONDS` environment variable (falls back to 60s on unset/invalid values).
2. **Multipathd executor timeout tied to `uxsock_timeout`** — a new package-level calculation in `linux/multipath.go` sets the multipathd command executor timeout to `uxsock_timeout + 5s` margin, so the Go executor never times out before multipathd's own socket timeout does.
3. **Skip multipath reconfigure when nothing changed** — `SetMultipathRecommendations` now hashes `multipath.conf` (SHA-256) and skips `multipathd reconfigure` when the applied hash matches, guarded by a cross-process `flock`. The applied hash is written atomically (temp file + `rename`).

Supporting changes:
- `mpathconfig/configresource.go`: map keys are now sorted before emitting config lines, making `multipath.conf` output **deterministic** across restarts (Go randomizes map iteration order). This is essential for the hash-based skip to work.
- `linux/device.go`: fixes a latent nil-pointer bug in `flushbufs` and routes `multipathd resize` through the new timeout-aware executor.
- `GetMultipathDevices` retry loop removed in favor of a single call with the longer timeout, returning `ErrMultipathTimeout` for callers to handle.
- Tests added/updated across `linux`, `tunelinux`, and `util`.

---

## Automated Scan Results

The `review_pr` tool flagged **2 critical "unchecked error return"** findings. Both are **false positives** — verified against the diff:

| Reported | Location | Verdict |
| --- | --- | --- |
| `if _, err = tempFile.WriteString(hash + "\n"); err != nil {` | `tunelinux/multipath.go:369` | ❌ False positive — the error **is** checked (`err != nil`) and the file is closed + error returned. |
| `if timeout, err := strconv.Atoi(val); err == nil && timeout > 0 {` | `util/cmd.go:45` | ❌ False positive — the error **is** checked (`err == nil`) as part of the guard. |

No genuine unchecked-error issues were found on manual inspection.

---

## Strengths

- **Correct concurrency handling.** The reconfigure-skip logic uses `flock (LOCK_EX)` to serialize across processes and writes the applied hash atomically via `CreateTemp` + `Sync` + `Rename`, avoiding torn reads.
- **Idempotent failure recovery.** The hash is written *only after* a successful `reconfigure()`. If reconfigure succeeds but the hash write fails, the next run simply reconfigures again — safe. Covered by `TestReconfigureMultipathdIfConfigNotAppliedRetriesAfterFailure`.
- **Deterministic config output.** Sorting map keys in `configresource.go` is the right dependency for hash stability — without it, the SHA would change every restart and defeat the optimization.
- **Real bug fix in `flushbufs`.** Changing `dev == nil && dev.AltFullPathName == ""` to `dev == nil || ...` fixes a nil-pointer dereference: the old `&&` form still evaluated `dev.AltFullPathName` when `dev == nil`, panicking. The `||` form short-circuits correctly.
- **Good test coverage** for the new pure logic: `loadMultipathdTimeouts` (default / configured / rounding / zero / negative / invalid), `loadDefaultTimeout`, and the reconfigure-skip state machine (apply once, skip on repeat, re-apply on change, read-error path).
- **Sensible timeout margin.** Rounding partial seconds up before adding the 5s margin guarantees the executor timeout always exceeds `uxsock_timeout`.

---

## Observations & Suggestions (non-blocking)

1. **Timeouts are cached at package init.** `multipathdUxsockTimeout/CommandTimeout` and `util.defaultTimeout` are resolved once via package-level `var` initialization. Runtime changes to the env vars are ignored (intentional, and asserted by `TestMultipathdCommandTimeoutIsCached`). Worth confirming the daemonset always sets these env vars *before* process start; document this so operators don't expect live reconfiguration.

2. **Behavior change: `GetMultipathDevices` no longer retries.** The previous 3-try/2s-sleep loop is replaced by a single attempt returning `ErrMultipathTimeout`. This relies on the new longer executor timeout (up to ~305s) absorbing transient slowness. Please confirm **all callers** treat `ErrMultipathTimeout` as "skip this cycle" rather than a hard failure, so a single slow `multipathd` response doesn't surface as an error to the user.

3. **Lock file lifecycle.** `appliedHashPath + ".lock"` is created but never removed. This is standard for lock files and harmless, but worth a one-line comment noting it is intentional.

4. **`MkdirAll(stateDir, 0755)` under `/host/run/...`.** Confirm the node container has write permission to the mounted host `run` path; a permission failure here would surface as an error from `SetMultipathRecommendations`. Consider whether that should be fatal or logged-and-skipped at the call site.

5. **Two env vars with different units.** `MULTIPATH_UXSOCK_TIMEOUT` is in **milliseconds** and `HPE_CMD_TIMEOUT_SECONDS` is in **seconds**. The multipathd command timeout derives from the former (+5s), not the latter. This is correct but easy to confuse — worth documenting in the daemonset/env reference.

---

## Knowledge Correlations (from MCP)

Related Jira / escalations surfaced by the scan (context, not defects):

- **Jira:** CON-4576, CON-4411, CON-4376, CON-4363, CON-4356, CON-4313, CON-2799, CON-1187, CON-778
- **Escalations:** 1322986, 1280439, 1278294, 1309464, 1298725

Several (CON-4376, CON-4363, CON-4411) relate to intermittent multipath establishment and NFS pod recreation timeouts — consistent with this PR's timeout-hardening direction.

---

## Recommendation

**Approve with minor comments.** The change is well-structured, correctly handles concurrency and atomicity, fixes a real nil-pointer bug, and is backed by focused unit tests. The two automated "critical" findings are false positives. Before merge, confirm the two behavioral items above: (a) all `GetMultipathDevices` callers tolerate `ErrMultipathTimeout` without surfacing a hard failure, and (b) the required env vars are set in the daemonset prior to process start.

rohit-balakrishna:
CSI_E2E_CONFIG=primera-iscsi-block-e2e

Ran 99 of 7907 Specs in 5075.229 seconds
PASS
Created /data/csi-e2e-nb2hr-26-08-07-07-57-16.xml.gz

rohit-balakrishna (approved):

dileepds (approved):

PR #238, opened 2026-08-05
CON-4887: ESC-17930: HPE NFS pods entered the Terminating state when one SAN network switch was shut down

rohit-balakrishna:


rohit-balakrishna:
CSI E2E Tests **(CSI_E2E_CONFIG=primera-iscsi-block-e2e** passed, results below:

Ran 99 of 7907 Specs in 4999.372 seconds
PASS
Created /data/csi-e2e-7bf9k-26-08-07-15-51-45.xml.gz

tunelinux/multipath.go
```diff
@@ -677,7 +679,9 @@ func RemoveMultipathDevice(multipathDevice string) error {
 	//check if the multipath device exists
 	log.Infof("Checking whether the multipath device %s exists or not.", multipathDevice)
 	if multipathDeviceExists(multipathDevice) {
-		_, _, err := util.ExecCommandOutput("dmsetup", []string{"remove", multipathDevice})
+		// Use multipath -f (not dmsetup remove) so multipathd stays in sync and kpartx/in-use
+		// maps are handled correctly.
+		_, _, err := util.ExecCommandOutput("multipath", []string{"-f", multipathDevice})
```
dileepds, line 684: Is this 'multipath' commands work always ?, with stale volume attachments as well ? As I remember 'multipath' command was not working always. We had couple of discussions on this when we implemented node monitor.

rohit-balakrishna: @dileepds , can you provide more details on what were the issues with multipath command? And why was dmsetup chosen?

rohit-balakrishna: @dileepds , as of now, node cleanup tears down multipath maps with a raw dmsetup remove, which operates at the low-level device-mapper layer and bypasses multipathd. Below is what the Copilot analysis is:

dmsetup acts directly on the kernel device-mapper table, bypassing the multipathd daemon. If multipathd is running, and if we run dmsetup remove mpatha.

- The kernel deletes the mapping.
- multipathd detects active physical paths (/dev/sdb, /dev/sdc) during its next path-check cycle.
- multipathd automatically re-creates /dev/mapper/mpatha.

To prevent this, use multipath -f mpatha instead, which updates the daemon state properly before flushing the map.

dileepds (approved):

PR #239, opened 2026-08-20
CON-4909 Updated NVME_DISCOVERY workflow and to eliminate any redundant discovery and connect calls

Yatish-Sriram:
CON-4909 Updated NVME_DISCOVERY workflow and to eliminate any redundant discovery and connect calls

### DIFF
--- a/linux/nvme.go
+++ b/linux/nvme.go
@@ -82,7 +82,13 @@ var nvmeCoreMultipathPath = "/sys/module/nvme_core/parameters/multipath"
 
 // nvmeExecCommandOutput runs an external command and returns its stdout, return
 // code, and error. It is a variable so unit tests can override it.
-var nvmeExecCommandOutput = util.ExecCommandOutput
+var (
+	nvmeExecCommandOutput       = util.ExecCommandOutput
+	nvmeControllerReadDir       = ioutil.ReadDir
+	nvmeControllerReadFirstLine = util.FileReadFirstLine
+	nvmeFindDevices             = FindNvmeDevices
+	nvmeApplyTcpTuning          = ApplyNvmeTcpTuning
+)
 
 // isNvmeMultipathEnabled reports whether native NVMe multipath is currently
 // enabled on this node by reading the read-only nvme_core.multipath module
@@ -134,10 +140,19 @@ func ConnectNvmeTarget(target *model.NvmeTarget) error {
 
 	var lastErr error
 	var success bool
+	var failedIPs []string
+
+	// Fix 3: skip portals that already have a live controller for this NQN.
+	liveAddrs, _ := getLiveNvmeControllersForNQN(target.NQN)
 
 	for _, ip := range discoveryIPs {
 		// Sanitize IP address (remove whitespace,* and other invalid format)
 		ip = util.SanitizeIPAddress(ip)
+		if liveAddrs[ip] {
+			log.Infof("NVMe target %s already connected via %s:%s, skipping redundant connect", target.NQN, ip, target.Port)
+			success = true
+			continue
+		}
 		args := []string{
 			"connect",
 			"-t", "tcp",
@@ -145,7 +160,7 @@ func ConnectNvmeTarget(target *model.NvmeTarget) error {
 			"-a", ip,
 			"-s", target.Port,
 		}
-		out, rc, err := util.ExecCommandOutput(nvmecmd, args)
+		out, rc, err := nvmeExecCommandOutput(nvmecmd, args)
 		// Treat rc=114 as success
 		if rc == 114 {
 			log.Infof("NVMe target %s connected via %s:%s (rc=114)", target.NQN, ip, target.Port)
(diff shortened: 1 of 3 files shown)


linux/nvme.go
```diff
@@ -134,18 +140,27 @@ func ConnectNvmeTarget(target *model.NvmeTarget) error {
 
 	var lastErr error
 	var success bool
+	var failedIPs []string
+
+	// Fix 3: skip portals that already have a live controller for this NQN.
+	liveAddrs, _ := getLiveNvmeControllersForNQN(target.NQN)
```
prashantkumars289: HandleNvmeTcpDiscovery >> ConnectNvmeTarget >> getLiveNvmeControllersForNQN
We are already calling the getLiveNvmeControllersForNQN function inside the HandleNvmeTcpDiscovery function. Do we need to call it here again? 

Passing the already computed liveAddrs into ConnectNvmeTarget would avoid a redundant '/sys/class/nvme' scan.

prashantkumars289:
Got a comment from Copilot review to add a test case for ConnectNvmeTarget when all portals are already live. Please address if necessary.

**No direct unit test for `ConnectNvmeTarget` when *all* portals are already live.** Existing tests cover "skip one live portal, connect the other" and the `HandleNvmeTcpDiscovery`-level full skip (which never calls `ConnectNvmeTarget` at all). A direct test asserting `ConnectNvmeTarget` returns success with zero `nvme connect` execs when every portal is live would close that gap for any other caller of `ConnectNvmeTarget`.

linux/nvme.go
```diff
@@ -134,18 +140,27 @@ func ConnectNvmeTarget(target *model.NvmeTarget) error {
...
 
 	for _, ip := range discoveryIPs {
```
rohit-balakrishna: should we rename this variable ? these IPs are not discovery IPs right?

linux/nvme.go
```diff
@@ -161,12 +176,20 @@ func ConnectNvmeTarget(target *model.NvmeTarget) error {
 			}
 			log.Warnf("NVMe connect failed for discovery IP %s, rc=%d, error: %v", ip, rc, err)
```
rohit-balakrishna, line 190: `discovery IP` is referred to here again.

linux/nvme.go
```diff
@@ -180,72 +203,122 @@ func RescanNvme() error {
...
-		if err := ConnectNvmeTarget(discTarget); err != nil {
-			return fmt.Errorf("failed to connect to NVMe target %s via discovered endpoints: %v", volume.Nqn, err)
-		}
+	log.Tracef("Using CSP-supplied NVMe/TCP target for NQN %s: address=%q, port=%q", target.NQN, target.Address, target.Port)
+
+	// Fix 2: skip nvme discover when a live controller for this NQN already
+	// exists, so the orphan-prone discovery controller is never created.
+	if len(liveAddrs) > 0 {
+		log.Infof("NVMe/TCP subsystem %s already connected via %v, skipping nvme discover", volume.Nqn, liveAddrs)
+	} else if discTarget := discoverNvmeTarget(volume); discTarget != nil {
```
rohit-balakrishna: Since, we already know the Target details, a discovery is not needed here. We can try to connect directly. That way, we will avoid the problem all together.

linux/nvme.go
Yatish-Sriram: Addressed with modified getLiveNvmeControllersForNQN() signature and passing pre-fetched live-address map.

linux/nvme.go
Yatish-Sriram: Addressed

linux/nvme.go
Yatish-Sriram: Addressed

linux/nvme.go
Yatish-Sriram: Addressed commented out the discovery workflow

Yatish-Sriram:
Removed NVME-DISCOVER operation from the NVME_STAGE_WORKFLOW and relying directly on NVME-CONNECT instead

rohit-balakrishna:
> Removed NVME-DISCOVER operation from the NVME_STAGE_WORKFLOW and relying directly on NVME-CONNECT instead

With the changes in the latest commit, we are no longer invoking discovery and hence the reap also is no longer necessary. As per the data shared by Dev(ERT), there are only 2 arrays that have created NVMe PVCs and both belong to this Escalation. Since we are not invoking the discovery now, a reap also is not necessary.

rohit-balakrishna (approved):

rohit-balakrishna:
CSI E2E passed with the latest commit.

[ReportAfterSuite] Kubernetes e2e JUnit report
k8s.io/kubernetes/test/e2e/framework/test_context.go:579
[ReportAfterSuite] PASSED [0.463 seconds]

Ran 99 of 7907 Specs in 4759.667 seconds
PASS
Created /data/csi-e2e-lb7m5-26-08-25-12-29-04.xml.gz

PR #249, opened 2026-09-16
CON-4955 AAPP hostSeesVLUN Block RWX at scale: WSAPI session exhaustion orphans primary VLUNs, and a publish-side peer-detail failure leaves 70% of volumes with no failover path and stranded secondary VLUNs

rashi-jain1:
JIRA Link: https://jira.storage.hpecorp.net/browse/CON-4955


linux/device.go
```diff
@@ -436,6 +436,18 @@ func rescanLoginVolume(volume *model.Volume) error {
...
+	// A replicated iSCSI volume whose CSP-supplied secondary payload carries no
+	// usable peer backends attaches with primary-only paths and no failover. Non-replicated
+	// volumes and genuine primary-failover both leave SecondaryArrayDetails empty, so this warns
+	// only for the true missing-peer case, which was previously silent.
+	if !strings.EqualFold(volume.AccessProtocol, "fc") &&
+		!strings.EqualFold(volume.AccessProtocol, nvmetcp) &&
+		volume.SecondaryArrayDetails != "" &&
+		len(secondaryBackends) == 0 {
+		log.Warnf("replicated volume %s came up with no secondary-array paths "+
+			"(primary-only, no failover); SecondaryArrayDetails=%q", volume.SerialNumber, volume.SecondaryArrayDetails)
```
rohit-balakrishna: @rashi-jain1 1 question, what is the content of volume.SecondaryArrayDetails ? Is it ok to dump the whole content of this variable?

rohit-balakrishna:
We can leave the bug number and trim the commit log to remove the reference to WSAPI and any content that is relevant only for the CSP. We can have the message that is relevant to CHL only. We can remove: "WSAPI session exhaustion orphans primary VLUNs, and a publish-side peer-detail failure leaves 70% of volumes "

rohit-balakrishna:
We need to understand and design on how to handle the degraded Volumes, where the secondary paths are missing. Can we raise an alert to notify, so that it is visible in the describe? This is not a comment for the current PR, but an overall comment.

rashi-jain1:
> We can leave the bug number and trim the commit log to remove the reference to WSAPI and any content that is relevant only for the CSP. We can have the message that is relevant to CHL only. We can remove: "WSAPI session exhaustion orphans primary VLUNs, and a publish-side peer-detail failure leaves 70% of volumes "

Bug number has already been removed and have edited commit msg

rashi-jain1:
> We need to understand and design on how to handle the degraded Volumes, where the secondary paths are missing. Can we raise an alert to notify, so that it is visible in the describe? This is not a comment for the current PR, but an overall comment.

Now we only add warn log when a repl volumes attaches primary only with no sec path For turning it into a proper alert that shows up in kubectl describe I think we need a separate design since it requires propagating the degraded state from node driver up to the CSP.

linux/device.go
rashi-jain1: volume.SecondaryArrayDetails is a JSON blob of the secondary array's per-LUN access info includes LUN IDs, target IQNs/NQNs, discovery IPs, and the iSCSI CHAP user/password. So dumping it would write the CHAP password to the node logs in cleartext. I've removed SecondaryArrayDetails=%q and kept just the volume serial

rohit-balakrishna:
Ran CSI-E2E test along with PR #248 and the results are below:
[ReportAfterSuite] Kubernetes e2e JUnit report
k8s.io/kubernetes/test/e2e/framework/test_context.go:579
[ReportAfterSuite] PASSED [0.411 seconds]

Ran 99 of 7907 Specs in 5201.820 seconds
PASS
Created /data/csi-e2e-4kvrn-26-09-22-09-23-17.xml.gz
rohitb@rb-esx-master:~/hpe-csi/deploy-yamls/nfs-provisioner-deploy$

rohit-balakrishna (approved):

